Issue #37, opened 2016-11-07
Use a standard RID structure in the nightly nupkg


Tratcher:
Right now the package has a completely non-standard layout, meaning it can be added to a project, but none of the binaries will be copied to the bin dir. If we want to test against the binaries we need a script to search the nuget package folder for the right dll path. Note it doesn't matter where the dll is on disk, so long as we have an absolute path to it that we can dynamically plug into the applicationhost.config template files for each test.

https://github.com/aspnet/AspNetCoreModule/issues/20, https://github.com/aspnet/ServerTests/issues/46

shirhatti:
FYI, for your reference https://github.com/dotnet/corefx/blob/master/pkg/Microsoft.NETCore.Platforms/runtime.json

moozzyk:
You can take a look at the libuv package - it has the right structure for shipping native libs in NuGet packages in the .NET Core world.

jhkimnew:
I have fixed this issue with updating nuget\AspNetCore.nuspec.

https://github.com/aspnet/AspNetCoreModule/commit/40ccbb3940b51ed380cddd6cc5867257818fd712
https://github.com/aspnet/AspNetCoreModule/commit/e119603cddbb57e1d09db054ef93ac71b4256ba8

Tratcher:
In hindsight I think this wasn't quite right. The rid structure is useful for keeping track of multiple binaries, but when we install/publish the package the tooling selects which bitness to use based on the application bitness, not the bitness of IIS/Express. There is no requirement that IIS/Express use the same bitness as the app, and in fact VS probably runs the x86 IIS Express so we want to make sure we test both versions. Right now we're only testing the x64 version.

New proposal: Use a package structure similar to the rid layout, but that preserves the structure when publishing so we can select the correct one at runtime based on which IIS Express we're going to launch. Can we deploy them as content?

jhkimnew:
Would you add some more information so that I can update the nuget package as you want?
What is the current outcome of the published result and what is expected result for the new proposal?

shirhatti:
I'm proposing create a folder structure rooted under the `win7` rid. This preserves both bitness versions of ANCM irrespective of the bitness of your app.

Does that sound reasonable to you folks?

Tratcher:
I don't know how to craft the package, but the published output I want is something like:
aspnetcoremodule/x86/aspnetcore.dll
aspnetcoremodule/x64/aspnetcore.dll

[...]

PR #42, opened 2016-11-17
Update nupkg layout to include both architectures together.

Tratcher:
https://github.com/aspnet/AspNetCoreModule/issues/37 @shirhatti 

@jhkimnew I think this layout change breaks your ps1.

### DIFF
--- a/nuget/AspNetCore.nuspec
+++ b/nuget/AspNetCore.nuspec
@@ -1,4 +1,4 @@
-<?xml version="1.0"?>
+﻿<?xml version="1.0"?>
 <package>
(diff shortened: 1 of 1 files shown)


jhkimnew (approved):
@Tratcher
This looks good to me. Please merge the change. 

@Tratcher @shirhatti @pranavkm 
One question. The installancm.ps1 seems not be used and it is out of dated. So, I think I should remove the file. Please let me know if you have any objection about that.

Tratcher:
@jhkimnew up to you, I don't need that script. @pan-wang?

Issue #44, opened 2016-11-18
w3wp crash issue on Win8.1 OS machine after applying the private build of aspnetcore.dll 


jhkimnew:
REPRO STEPS:

1. Prepare Windows 8.1 and install IIS and .NetCore bundle
2. Publish an aspnetcore app to the Default Web Site
3. Clone this repository and build aspnetcore.dll for both x86 and amd64 
4. Replace the aspnetcore.dll on %windir%\system32\inetsrv\aspnetcore.dll and %windir%\syswow64\inetsrv\aspnetcore.dll 
5. Enable the 32 bit mode of DefaultAppPool
6. Send a request to the Default Web Site
7. Recycle DefaultAppPool

EXPECTED:

No crashe

ACTUAL:

W3wp.exe crashes.
If we enable appverifier and attach windbg to the w3wp process, we can get more information about the crash. 
Here is the error message I got from the appverifier and stack information.
 
<><><>

0:048> g

=======================================
VERIFIER STOP 00000006: pid 0x28C: Corrupted heap pointer or using wrong heap. 

                FE2B1000 : Heap handle used in the call.
                EF5F0A38 : Heap block involved in the operation.
                000005C8 : Size of the heap block.
                FBF91000 : Heap where block was originally allocated.

=======================================
This verifier stop is not continuable. Process will be terminated 
when you use the `go' debugger command.

=======================================

(28c.11ac): WOW64 breakpoint - code 4000001f (first chance)
First chance exceptions are reported before any exception handling.
This exception may be expected and handled.
vrfcore!VerifierStopMessageEx+0x5a3:
570a3ba3 cc              int     3

0:004:x86> kn
# ChildEBP RetAddr  
00 043cf4f4 570a87a8 vrfcore!VerifierStopMessageEx+0x5a3 d:\rs2\onecore\base\avrf\avrf30\vrfcore\sdk.cpp @ 586]
01 043cf51c 590cc434 vrfcore!VfCoreRedirectedStopMessage+0x88 d:\rs2\onecore\base\avrf\avrf30\vrfcore\stopredirect.cpp @ 85]
02 043cf570 590ca455 VERIFIER_590c0000!VerifierStopMessage+0x74 d:\9147\base\avrf\verifier\logging.c @ 221]
03 043cf5d4 590ca721 VERIFIER_590c0000!AVrfpDphReportCorruptedBlock+0x185 d:\9147\base\avrf\verifier\heappage.c @ 6035]
[...]

PR #56, opened 2017-01-06
Fixing issue #44

jhkimnew:
@pan-wang @shirhatti  Please review this and approve so that I can merge this bug fix.
Two files were updated, one for bug fix and the other for .gitignore to the change of AspNetCoreModule.VC.db file.


pan-wang:
acked

PR #68, opened 2017-02-09
Remove MAX_PATH buffer length to support long file path

jhkimnew:
@pan-wang @shirhatti Please review this PR and approve this

### DIFF
--- a/src/AspNetCore/Src/path.cxx
+++ b/src/AspNetCore/Src/path.cxx
@@ -341,6 +341,7 @@ PATH::IsValidHeaderName(
     return TRUE;
 }
 
+ 
 HRESULT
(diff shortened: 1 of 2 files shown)


pan-wang:
looks good

shirhatti:
I just noticed this PR. I'm assuming the only changes to `serverprocess.cxx` were to do with whitespaces?

jhkimnew:
@shirhatti @pan-wang 

Serverproce.cxx has the right fix. However, there was a mistake with handling the white space because I used both Visual Studio and Notepad2.exe editor and TAB key seems to be converted to space or reverse change. I will try to roll back the white space change so that the actual code change visible.

PR #85, opened 2017-04-03
Update Korebuild and dependency versions

shirhatti:


jhkimnew:
Looks good.

natemcmaster:
You should also delete makefile.shade. It's not used

PR #98, opened 2017-05-16
Added ANCMTestFlags environment variable

jhkimnew:
@shirhatti @pan-wang 

Added a new environment variable.
If you want to skip ANCM test, you can run below commands before executing ".\build.cmd".

    Set ANCMTestFlags=SkipTest
    or
    $env:ANCMTestFlags="SkipTest"


pan-wang (approved):
looks good

PR #100, opened 2017-05-17
Disable using privateAspNetCoreFile to test release candiate build

jhkimnew:
@shirhatti  @pan-wang 
Because CI machine has newer ANCM build, I disabled the test option to use the private aspnetcore.dll. This will be back when we sync aspnetcore.dll source files with the latest version.


Tratcher:
Should this be changed to an environment variable so we don't have to keep changing the code? The variable can be set on the CI.

jhkimnew:
Currently I don't have permission to add or set any environment variable on the CI. Once I get the permission, I will replace it with environment variable.

PR #102, opened 2017-05-18
add features: flowing authentication info, hosting environment variab…

pan-wang:
…le support; fix client disconnect and app_offline issues


jhkimnew:
@pan-wang 
Looks good. I confirmed that there is no build issue on my dev machine after checking out this private branch.

jhkimnew (approved):

PR #110, opened 2017-06-12
Change nuget pacakge version to include -pre-

Tratcher:
Works around a nuget bug with numeric pre-release versions: https://github.com/NuGet/Home/issues/4513


jhkimnew (approved):

CodingSpiderFox:
When is this going to be released?

PR #126, opened 2017-08-01
[WIP] Native IIS modifications

jkotalik:
where my progress will be

### DIFF
--- a/samples/SampleServer/IISHttpServer.cs
+++ b/samples/SampleServer/IISHttpServer.cs
@@ -14,7 +14,7 @@ namespace SampleServer
     {
         private static NativeMethods.PFN_REQUEST_HANDLER _requestHandler = HandleRequest;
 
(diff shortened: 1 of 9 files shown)


samples/SampleServer/IISHttpServer.cs
```diff
@@ -50,19 +48,32 @@ public Task StopAsync(CancellationToken cancellationToken)
...
+            var task = server.ProcessRequest(context);
 
-            // TODO: Handle sync completion
-            return NativeMethods.REQUEST_NOTIFICATION_STATUS.RQ_NOTIFICATION_PENDING;
+            var awaiter = task.GetAwaiter();
+
+            if (awaiter.IsCompleted)
+            {
+                // Synchronous
+                awaiter.GetResult();
```
davidfowl: context.ProcessRequest still posts a completion in the sync case, we should fix that.

PR #129, opened 2017-08-04
fix websocket connection issue and some memory leak, and add debug print

pan-wang:
@shirhatti @jhkimnew could you please review the change and approve the change


jhkimnew (approved):
Looks good. Great job!!!

PR #131, opened 2017-08-10
Sortable log files with stdoutLogFile

### LINKED ISSUE #130: Sortable log files with stdoutLogFile
Opening a new issue, since [the previous one was closed without a proper solution and ignoring the problem](https://github.com/aspnet/AspNetCoreModule/issues/103).

The stdout logging is awesome and it works great.
However, when I look in my App Service folder, I see the following. 🤔 

```
> ls logs
D:\home\site\wwwroot
stdout_10652_20178414246.log
stdout_316_20178415450.log
stdout_6504_201783143517.log
stdout_8068_20178525714.log
stdout_9224_201784122443.log
```

I challenge you to try to find the latest file. 😉 (my personal record stands at 15 sec)
I'm already struggling to read timestamps with just 5 files. This becomes *utterly unmanageable* over time.

- The first digit seems to be the process ID (PID) as it doesn't match the port the server is binding to?
- And the timestamp uses a *short* date and time (?!) format. (not sure if date or time digit? 😨)

**I suggest to change to the following format: *prefix*\_*date*\_*pid***

Where the date uses the [format](https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings) yyyyMMddhhmmss.

Example of the same files with the suggested format:

```
> ls logs
D:\home\site\wwwroot
stdout_20170803143517_6504.log
stdout_20170804122443_9224.log
stdout_20170804140246_10652.log
stdout_20170804150450_316.log
stdout_20170805025714_8068.log
```

Much better! 😄
And now it's sortable I can spot that on the 4th there were 3 files created and in what order. Useful!

Oh, and please document this filename behaviour on this page: https://docs.microsoft.com/en-us/aspnet/core/hosting/aspnet-core-module

For reference, this is my Web.config file:

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <system.webServer>
    <handlers>
      <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified"/>
    </handlers>
    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="true" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false"/>
  </system.webServer>
</configuration>
```

CC: @glatzert

guardrex:
Fixes #130 

I know u didn't label it *up-for-grabs*, but I just wanted to offer it ... plz feel free to close if unwelcome or poorly timed (i.e., if it's bad time for a breaking change).

cc/ @shirhatti @pan-wang

### DIFF
--- a/src/AspNetCore/Src/serverprocess.cxx
+++ b/src/AspNetCore/Src/serverprocess.cxx
@@ -1204,15 +1204,15 @@ SERVER_PROCESS::SetupStdHandles(
         }
 
         GetSystemTime(&systemTime);
-        hr = struLogFileName.SafeSnwprintf(L"%s_%d_%d%d%d%d%d%d.log",
+        hr = struLogFileName.SafeSnwprintf(L"%s_%d%02d%02d%02d%02d%02d_%d.log",
                                            struAbsLogFilePath.QueryStr(),
-                                           GetCurrentProcessId(),
                                            systemTime.wYear,
                                            systemTime.wMonth,
                                            systemTime.wDay,
                                            systemTime.wHour,
                                            systemTime.wMinute,
-                                           systemTime.wSecond );
+                                           systemTime.wSecond,
+                                           GetCurrentProcessId() );
         if (FAILED(hr))
         {
             goto Finished;


bgever:
Awesome, thanks @guardrex!

Not likely to break much, as the date format wasn't really possible to process anyway due to its short form. One possible breaking part is the swap of timestamp and PID, if someone was expecting this specific order. But my gut feeling is nobody was parsing these filenames anyway. ;-)

PR #132, opened 2017-08-10
Uses content path from ancm and full path for dll for dotnet.exe

jkotalik:
@davidfowl


src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -10,6 +10,10 @@ extern "C" __declspec(dllexport) void register_request_callback(PFN_REQUEST_HAND
     ASPNETCORE_APPLICATION::GetInstance()->SetRequestHandlerCallback(requestHandler, pvRequstHandlerContext);
 }
 
+//extern "C" __declspec(dllexport) PCWSTR get_application_path(IHttpContext* pHttpContext)
```
davidfowl: remove

davidfowl (approved):

PR #134, opened 2017-08-15
[WIP] Initial shutdown logic. 

jkotalik:
So you have an idea of what is going on. We may need to provide a good way to reference the running application to shutdown the managed code.

### DIFF
--- a/samples/SampleServer/IISHttpServer.cs
+++ b/samples/SampleServer/IISHttpServer.cs
@@ -13,13 +13,20 @@ namespace SampleServer
     public class IISHttpServer : IServer
     {
         private static NativeMethods.PFN_REQUEST_HANDLER _requestHandler = HandleRequest;
+        private static NativeMethods.PFN_SHUTDOWN_HANDLER _shutdownHandler = HandleShutdown;
+
 
         private IISContextFactory _iisContextFactory;
 
         private PipeFactory _pipeFactory = new PipeFactory();
         private GCHandle _httpServerHandle;
+        private IApplicationLifetime _applicationLifetime;
 
         public IFeatureCollection Features { get; } = new FeatureCollection();
+        public IISHttpServer(IApplicationLifetime applicationLifetime)
+        {
+            _applicationLifetime = applicationLifetime;
+        }
 
         public Task StartAsync<TContext>(IHttpApplication<TContext> application, CancellationToken cancellationToken)
         {
@@ -28,7 +35,8 @@ namespace SampleServer
             _iisContextFactory = new IISContextFactory<TContext>(_pipeFactory, application);
 
             // Start the server by registering the callback
-            NativeMethods.register_request_callback(_requestHandler, (IntPtr)_httpServerHandle);
+            // TODO the context may change here for shutdown.
+            NativeMethods.register_callbacks(_requestHandler, _shutdownHandler, (IntPtr)_httpServerHandle, (IntPtr)_httpServerHandle);
 
             return Task.CompletedTask;
         }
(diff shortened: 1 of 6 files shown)


samples/SampleServer/IISHttpServer.cs
```diff
@@ -73,6 +76,14 @@ private static NativeMethods.REQUEST_NOTIFICATION_STATUS HandleRequest(IntPtr pH
             return NativeMethods.REQUEST_NOTIFICATION_STATUS.RQ_NOTIFICATION_PENDING;
         }
 
+        private static NativeMethods.REQUEST_NOTIFICATION_STATUS HandleShutdown(IntPtr pHttpContext, IntPtr pvRequestContext)
+        {
+            // TODO handle shutting down the application. 
+            var server = (IISHttpServer)GCHandle.FromIntPtr(pvRequestContext).Target;
+            server.Dispose();
+            return NativeMethods.REQUEST_NOTIFICATION_STATUS.RQ_NOTIFICATION_FINISH_REQUEST;
```
jkotalik: This probably should just return a bool instead.

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -348,5 +362,8 @@ REQUEST_NOTIFICATION_STATUS ASPNETCORE_APPLICATION::ExecuteRequest(IHttpContext*
 
 void ASPNETCORE_APPLICATION::Shutdown()
 {
-
+    // First call into the managed server and shutdown
+
```
jkotalik: TODO is there anything else we would need to clear here?

davidfowl: The static maybe?

src/AspNetCore/Inc/aspnetcoreapplication.h
```diff
@@ -14,6 +15,7 @@ class ASPNETCORE_APPLICATION
     REQUEST_NOTIFICATION_STATUS ExecuteRequest(IHttpContext* pHttpContext);
     void Shutdown();
     void SetRequestHandlerCallback(PFN_REQUEST_HANDLER callback, void* pvRequstHandlerContext);
+    void SetShutdownHandlerCallback(PFN_REQUEST_HANDLER callback, void* pvRequstHandlerContext);
```
davidfowl: Lets make a single initialize that passes all of the delegates.

src/AspNetCore/Src/aspnetcoreconfig.cxx
```diff
@@ -8,16 +8,21 @@ ASPNETCORE_CONFIG::~ASPNETCORE_CONFIG()
     //
     // the destructor will be called once IIS decides to recycle the module context (i.e., application)
     //
+    // shutting down core application first
+    if (ASPNETCORE_APPLICATION::GetInstance() != NULL) {
```
davidfowl, line 12: This might not be right. Ideally we should be called in response to RecycleApplication being called. So the flow is basically:

RecycleApplication -> Shutdown ASP.NET Core.

Are you sure this is the place that triggers shutdown of dotnet.exe with regular ANCM?

jkotalik: I'm not sure about if this is where dotnet.exe is shutdown, but it is triggered by the app pool being closed on IIS.

davidfowl: I see. It might be an ungraceful shutdown.

samples/SampleServer/IISHttpServer.cs
```diff
@@ -73,6 +76,14 @@ private static NativeMethods.REQUEST_NOTIFICATION_STATUS HandleRequest(IntPtr pH
             return NativeMethods.REQUEST_NOTIFICATION_STATUS.RQ_NOTIFICATION_PENDING;
         }
 
+        private static bool HandleShutdown(IntPtr pvRequestContext)
+        {
+            // TODO handle shutting do
+            var server = (IISHttpServer)GCHandle.FromIntPtr(pvRequestContext).Target;
+            server.Dispose();
```
davidfowl: This needs to shut the application down, not the server. The IISHttpServer should take an `IApplicationLifetime` in the constructor and this code would call StopApplication as a result.

jkotalik:
@davidfowl cool if I merge this?

PR #141, opened 2017-08-30
[WIP] Fixing win7 compilation issues.

jkotalik:
I verified that this change works on win7 and win8 with web sockets. Further verification is of course needed, but I wanted to get some light on this fix. 
@anurse

### DIFF
--- a/src/AspNetCore/Src/precomp.hxx
+++ b/src/AspNetCore/Src/precomp.hxx
@@ -7,9 +7,12 @@
 //
 // System related headers
 //
-#define NTDDI_VERSION 0x06020000
-#define _WIN32_WINNT 0x0602
 #define _WINSOCKAPI_
+
+#define NTDDI_VERSION 0x06010000
+#define WINVER 0x0601
+#define _WIN32_WINNT 0x0601
+
 #include <windows.h>
 #include <atlbase.h>
 #include <pdh.h>
@@ -17,7 +20,26 @@
 //#include <ntassert.h>
 #include <Shlobj.h>
 #include <httpserv.h>
+
+// So this "hack" should remove our issue of compiling for win7 without header files.
+// We will force ourselves to believe we are on win8 for the iiswebsocket.h file
+// and then assume we are compiling for win7 afterwards.
+#undef NTDDI_VERSION
+#undef WINVER
+#undef _WIN32_WINNT
+#define NTDDI_VERSION 0x06020000
+#define WINVER 0x0602
+#define _WIN32_WINNT 0x0602
 #include <iiswebsocket.h>
+#undef NTDDI_VERSION
+#undef WINVER
+#undef _WIN32_WINNT
+
+#define NTDDI_VERSION 0x06010000
+#define WINVER 0x0601
+#define _WIN32_WINNT 0x0601
+
+//#include <ntassert.h>
 #include <httptrace.h>
 #include <winhttp.h>
 
--- a/src/AspNetCore/Src/precomp.hxx
+++ b/src/AspNetCore/Src/precomp.hxx
@@ -21,9 +21,9 @@
 #include <Shlobj.h>
 #include <httpserv.h>
 
-// So this "hack" should remove our issue of compiling for win7 without header files.
-// We will force ourselves to believe we are on win8 for the iiswebsocket.h file
-// and then assume we are compiling for win7 afterwards.
+// This should remove our issue of compiling for win7 without header files.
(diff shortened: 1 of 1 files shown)


src/AspNetCore/Src/precomp.hxx
```diff
@@ -7,17 +7,39 @@
...
+
 #include <windows.h>
 #include <atlbase.h>
 #include <pdh.h>
 
 //#include <ntassert.h>
 #include <Shlobj.h>
 #include <httpserv.h>
+
+// So this "hack" should remove our issue of compiling for win7 without header files.
```
Tratcher: policheck... remove `hack`

src/AspNetCore/Src/precomp.hxx
```diff
@@ -7,17 +7,39 @@
...
 #include <iiswebsocket.h>
+#undef NTDDI_VERSION
+#undef WINVER
+#undef _WIN32_WINNT
+
+#define NTDDI_VERSION 0x06010000
+#define WINVER 0x0601
+#define _WIN32_WINNT 0x0601
+
+//#include <ntassert.h>
```
Tratcher: extra?

src/AspNetCore/Src/precomp.hxx
```diff
@@ -7,17 +7,39 @@
...
 #include <windows.h>
 #include <atlbase.h>
 #include <pdh.h>
 
 //#include <ntassert.h>
 #include <Shlobj.h>
 #include <httpserv.h>
+
+// So this "hack" should remove our issue of compiling for win7 without header files.
+// We will force ourselves to believe we are on win8 for the iiswebsocket.h file
```
analogrelay: I would rephrase slightly as "We force the Windows 8 version check logic in iiswebsocket.h to succeed even though we're compiling for Windows 7. Then set the version defines back to Windows 7 for the rest of the compilation"

NBD, but because of the magic here we want a nice clear comment :).

analogrelay (approved):
I love it! Hack it until it works!

analogrelay:
This is awesome. Now that we've got a proof-of-concept that we can build the GitHub version and run in both Win7 and Win8, we can look at migrating the source-of-truth to GitHub. /cc @davidfowl

jkotalik:
@anurse Is there a way to store what the previous defined value of NTDDI_VERSION? If so, we should probably restore it to that, rather than forcing Windows 7 afterwards.

analogrelay:
I'm not sure it's worth investing a lot in that. The original values are set up at the top of the file already, so it's not like it's hard to get them. It was already forced to Windows 7 at the start, you're just manually restoring the values.

If you wanted to, you could try using the [`push_macro`/`pop_macro` compiler pragmas](https://msdn.microsoft.com/en-us/library/hsttss76.aspx) to save the original values and then restore them (rather than hard-coding the original value a second time). That would mean it would only compile in VC++, but it's not like we plan to use Clang/GCC for ANCM anytime soon ;) (plus I think those compilers support it as well, in the name of broad compiler compatibility).

It may be possible to rearrange this file so that we don't do the weird dance of: set to windows 7, override to windows 8, override back to windows 7. Instead, we could set to windows 8 at the start, include `iiswebsocket.h` and then set to windows 7 for the rest of the file (so we aren't flipping back and forth), but `#include`s are order-sensitive so it's possible there's a problem with doing that.

analogrelay:
Also, we change three `#define`s, but it's possible only one or two of them are required. Might be worth doing a little differential testing (and looking at `iiswebsocket.h`) to find out exactly which `#define`s are actually required and limit the change to those.

jkotalik:
@jhkimnew Please run tests for ANCM with this change.

jhkimnew:
@pan-wang would you check if we can allow this change? If Pan agrees with this change, I will test this on Win7.

jkotalik:
@pan-wang and @shirhatti agreed with me offline (for some reason Pan can't comment on this issue). Test this both on Win7 and Win8+websockets if possible.

pan-wang:
The change looks good to me. We need to run tests to make sure there is no regression





On 2017-09-01 11:33 , Justin Kotalik<mailto:notifications@github.com> Wrote:


@pan-wang<https://github.com/pan-wang> and @shirhatti<https://github.com/shirhatti> agreed with me offline (for some reason Pan can't comment on this issue). Test this both on Win7 and Win8+websockets if possible.

shirhatti (approved):

jkotalik:
@jhkimnew Have you gotten the chance to test these changes?

jkotalik:
@jhkimnew said offline that he believes the change is good to go.

PR #144, opened 2017-09-01
Adds ANCM in-process flag.

jkotalik:
Name is up for discussion. @shirhatti I decided a bool is better for now.

### DIFF
--- a/src/AspNetCore/aspnetcore_schema.xml
+++ b/src/AspNetCore/aspnetcore_schema.xml
@@ -25,7 +25,7 @@
         <attribute name="processesPerApplication" type="uint" defaultValue="1" validationType="integerRange" validationParameter="1,100"/>
         <attribute name="forwardWindowsAuthToken" type="bool" defaultValue="true" />
         <attribute name="disableStartUpErrorPage" type="bool" defaultValue="false" />
-        <attribute name="hostingMode" type="string" defaultValue="outofprocess" />
+        <attribute name="hostingModel" type="string" defaultValue="outofprocess" />
         <element name="recycleOnFileChange">
             <collection addElement="file" clearElement="clear">
                 <attribute name="path" type="string" required="true" validationType="nonEmptyString" expanded="true"/>
--- a/src/AspNetCore/aspnetcore_schema.xml
+++ b/src/AspNetCore/aspnetcore_schema.xml
@@ -25,6 +25,7 @@
         <attribute name="processesPerApplication" type="uint" defaultValue="1" validationType="integerRange" validationParameter="1,100"/>
         <attribute name="forwardWindowsAuthToken" type="bool" defaultValue="true" />
         <attribute name="disableStartUpErrorPage" type="bool" defaultValue="false" />
+        <attribute name="applicationInProcess" type="bool" defaultValue="false" />
         <element name="recycleOnFileChange">
             <collection addElement="file" clearElement="clear">
                 <attribute name="path" type="string" required="true" validationType="nonEmptyString" expanded="true"/>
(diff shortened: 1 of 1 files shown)


Tratcher:
Not a fan, it's not extensible. How about hostingMode="inprocess" or "outofprocess"?

muratg:
@Tratcher What kind of extensibility do you envision?  Do we expect to have other kinds of hosting modes in the future?

We had a talk with @shirhatti about ideally making this the last web.config schema change for ANCM, and possibly introduce other ways to configure it (e.g. `ANCM.json`).  We'd like to get to a point that all ANCM updates could be done via a siteextension.

Tratcher:
Extensibilty such as hostingMode="thenextgreatthing", which would not require a schema change?

shirhatti:
cc @DamianEdwards

pan-wang:
with in-proc, do we still plan to support other configuration elements, such as environment variables?

Tratcher:
We should attempt to support as many as possible. The transition between the two modes should be as seamless as possible.

muratg:
Agreed with @Tratcher.  Are there any we can't support?

davidfowl:
Yes, we should try to support everything.

davidfowl:
I was thinking about this a bit and I like the string thing. You can imagine the string representing the transport type:

- "http"
- "inprocess"
- "namedpipes"
- "fastcgi"

Tratcher:
Also, "https", or "http://ip:port" for hard coding the forward ip and port? Some people really don't like our random port mechanics.

Tratcher:
Maybe it should be named `transport` instead.

Tratcher:
hmm, no, transport makes more sense as a separate thing from process model.

davidfowl:
@Tratcher I don't care if we call it transport or not I think the same knob should be used if we choose to change the communication mechanism between ANCM and the managed process.

Tratcher:
Disagree. How you launch the process is rather independent of how you connect to it. The inproc case is the exception. E.g. Ways to launch the process: inproc, outofproc, none (managed externally), etc.

pan-wang:
I prefer hostingModel or processModel. Transport is so misleading for in-proc or out-proc.

davidfowl:
So we'll do "inprocess" "http" that doesn't make any sense. What's the scenario where having separate options makes sense? Give me something real...

Once again call it mode, I just don't think there's just inproc and outproc. It's about how the module communicates with dotnet. That's what this mode represents

Tratcher:
Not every combo will make sense, but consider the following pairs:
outofproc + http
outofproc + https
outofproc + http2
outofproc + namedpipe
cluster + http
cluster + https
cluster + http2
cluster + namedpipe
selfstarted + http
selfstarted + https
selfstarted + http2
selfstarted + namedpipe
inproc + direct
inproc + namedpipe

davidfowl:
@Tratcher you're making things up now. What is inproc named pipe and inproc direct?

davidfowl:
To summarize my position, this new communication "mode" is how ANCM talks to dotnet. Reasons:

1. It’s simpler
2. Nothing is broken
3. No need to invalidate bad combinations (like inproc + named pipes)

The purity argument that they are different things requires that we explode or ignore when you have an invalid combination (like in process + http).

jkotalik:
Based on the feedback, I'm going to change the name to hostingModel and the type to a string. @Tratcher we may want to expose transport eventually, but I think that could be added to some eventual `ANCM.json` file.

src/AspNetCore/aspnetcore_schema.xml
```diff
@@ -25,6 +25,7 @@
         <attribute name="processesPerApplication" type="uint" defaultValue="1" validationType="integerRange" validationParameter="1,100"/>
         <attribute name="forwardWindowsAuthToken" type="bool" defaultValue="true" />
         <attribute name="disableStartUpErrorPage" type="bool" defaultValue="false" />
+        <attribute name="hostingModel" type="string" defaultValue="outofprocess" />
```
Tratcher: can you not have a default? e.g. default to empty? Let's handle the default in code.

pan-wang:
I understand the motivation that you want to make the new string attribute as a generic solution and thus no default value given, though I don't like it as it looks we don't know what possible values could be.

pan-wang (approved):

pan-wang:
@jkotalik you can merge this schema change now now

davidfowl:
default should be http

PR #151, opened 2017-09-11
Fix websocket close handshake issue and race condition when websocket…

pan-wang:
… client disconnect without close handshake #77

### DIFF
--- a/src/AspNetCore/Src/aspnetcoreconfig.cxx
+++ b/src/AspNetCore/Src/aspnetcoreconfig.cxx
@@ -376,6 +376,7 @@ ASPNETCORE_CONFIG::Populate(
         strExpandedEnvValue.Reset();
         pEnvVar->Release();
         pEnvVar = NULL;
+        pEntry->Dereference();
         pEntry = NULL;
     }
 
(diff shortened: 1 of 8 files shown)


src/AspNetCore/Src/aspnetcoreconfig.cxx
```diff
@@ -376,6 +376,7 @@ ASPNETCORE_CONFIG::Populate(
         strExpandedEnvValue.Reset();
         pEnvVar->Release();
         pEnvVar = NULL;
+        pEntry->Dereference();
```
jkotalik, line 379: Is this related to the fix or is it unrelated.

pan-wang: this fixes a memory leak

jhkimnew (approved):
Looks good to me.

pan-wang:
thanks

risogolo:
So, when this bugfix will be released as a part of new version of nuget package for websockets?

Tratcher:
@risogolo the bugfix is not in the websocket nuget package, it's in the AspNetCoreModule which installs via MSI. Expect the MSI with 2.1.0-preview.

risogolo:
but this bug is also in 1.x

Tratcher:
ANCM does not version like packages, it's an in place update. The next MSI released will affect all versions of ASP.NET Core.

PR #152, opened 2017-09-11
Adds in-process mode to ANCM.

jkotalik:
This is what is currently on our prototype branch for supporting the in-process mode. Wanted to expose what needs to be added to ANCM.

Work items that are still left:
- [x] Removing "wstring" (mostly)
- [x] <del>Port changes to proxy module to a separate in-process module</del> Can't do because we won't have the aspnetcore config early enough
- [ ] Tests and verification
- [x] Finding hostfxr logic (currently copies from core-setup)

There is more work that will be occurring in [aspnet/IISIntegration](https://github.com/aspnet/IISIntegration/tree/jkotalik/inprocess)
@davidfowl @pan-wang @shirhatti


AspNetCoreModule.sln
```diff
@@ -34,20 +34,20 @@ Global
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
 		{439824F9-1455-4CC4-BD79-B44FA0A16552}.Debug|Any CPU.ActiveCfg = Debug|Win32
-		{439824F9-1455-4CC4-BD79-B44FA0A16552}.Debug|Win32.ActiveCfg = Release|x64
-		{439824F9-1455-4CC4-BD79-B44FA0A16552}.Debug|Win32.Build.0 = Release|x64
-		{439824F9-1455-4CC4-BD79-B44FA0A16552}.Debug|x64.ActiveCfg = Release|x64
-		{439824F9-1455-4CC4-BD79-B44FA0A16552}.Debug|x64.Build.0 = Release|x64
+		{439824F9-1455-4CC4-BD79-B44FA0A16552}.Debug|Win32.ActiveCfg = Debug|x64
```
pan-wang: why do you want to remove release build

src/AspNetCore/Inc/application.h
```diff
@@ -181,6 +183,47 @@ class APPLICATION
...
+        BOOL fLockTaken = FALSE;
+
+        if (m_pAspNetCoreApplication == NULL)
+        {
+            AcquireSRWLockExclusive(&m_srwLock);
+            fLockTaken = TRUE;
+
+            if (m_pAspNetCoreApplication == NULL)
+            {
+                auto application = new ASPNETCORE_APPLICATION();
```
pan-wang: please do null check on the returned pointer  to avoid AV

src/AspNetCore/Inc/application.h
```diff
@@ -181,6 +183,47 @@ class APPLICATION
...
+                }
+
+                // Assign after initialization
+                m_pAspNetCoreApplication = application;
+            }
+        }
+
+        *ppAspNetCoreApplication = m_pAspNetCoreApplication;
+
+    Failed:
```
pan-wang: it's better to name it as Finished instead of Failed here

src/AspNetCore/Inc/application.h
```diff
@@ -226,14 +269,16 @@ class APPLICATION
...
+    STRU                    m_strAppPhysicalPath;
+    mutable LONG            m_cRefs;
+    APPLICATION_KEY         m_applicationKey;
+    PROCESS_MANAGER*        m_pProcessManager;
+    APPLICATION_MANAGER    *m_pApplicationManager;
+    BOOL                    m_fAppOfflineFound;
+    APP_OFFLINE_HTM        *m_pAppOfflineHtm;
+    FILE_WATCHER_ENTRY     *m_pFileWatcherEntry;
+    ASPNETCORE_APPLICATION *m_pAspNetCoreApplication;
+    SRWLOCK                m_srwLock;
```
pan-wang: need extra white space here

src/AspNetCore/Inc/aspnetcoreapplication.h
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+    static ASPNETCORE_APPLICATION* GetInstance()
+    {
+        return s_Application;
+    }
+
+private:
+    // Thread executing the .NET Core process
+    HANDLE m_Thread;
```
pan-wang: to keep consistent with other code, please rename it as m_hThread

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,359 @@
...
+    // Initialization complete
+    SetEvent(m_InitalizeEvent);
+}
+
+
+
+HRESULT ASPNETCORE_APPLICATION::Initialize(ASPNETCORE_CONFIG * pConfig)
+{
+    m_pConfiguration = pConfig;
+    m_InitalizeEvent = CreateEvent(
```
pan-wang: what does it mean if CreateEvent failed

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,359 @@
...
+        0,          // default creation flags
+        NULL);      // receive thread identifier
+
+    if (m_Thread == NULL)
+    {
+        return HRESULT_FROM_WIN32(GetLastError());
+    }
+
+    // If the debugger is attached, never timeout
+    auto dwTimeout = INFINITE; // IsDebuggerPresent() ? INFINITE : pConfig->QueryStartupTimeLimitInMS();
```
pan-wang: you need to update it to meet the goal

pan-wang: I am not a fun of auto. I know IIS code is not using a lot of new staff

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,359 @@
...
+	auto length = ::GetEnvironmentVariableW(name, nullptr, 0);
+	if (length == 0)
+	{
+		auto err = GetLastError();
+		if (err != ERROR_ENVVAR_NOT_FOUND)
+		{
+		}
+		return false;
+	}
+	auto buf = new wchar_t[length];
```
pan-wang: please do null check

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,359 @@
...
+
+    // We set a static so that managed code can call back into this instance and
+    // set the callbacks
+    s_Application = this;
+
+    m_ProcessExitCode = proc(2, argv);
+}
+
+BOOL ASPNETCORE_APPLICATION::GetEnv(const wchar_t* name, std::wstring *recv) {
+	recv->clear();
```
pan-wang: why call recv->clean

pan-wang: this is a public method, you may want to do null check on name and recv, at least add debug assert to make sure no null value is allowed

jkotalik:
@pan-wang I'm going to convert aspnetcoreapplication.cxx to use STRU next, so hold off on commenting on ExecuteApplication().

src/AspNetCore/Inc/fx_ver.h
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#ifndef __FX_VER_H__
```
davidfowl, line 4: /cc @Eilon this is copied from dotnet/core-setup

src/AspNetCore/Src/proxymodule.cxx
```diff
@@ -79,32 +79,82 @@ CProxyModule::OnExecuteRequestHandler(
...
-        pHttpContext->GetResponse()->SetStatus(500, "Internal Server Error", 0, E_OUTOFMEMORY);
+    HRESULT hr;
+    APPLICATION_MANAGER* pApplicationManager;
+    APPLICATION* pApplication;
+    ASPNETCORE_CONFIG* config;
+    ASPNETCORE_APPLICATION* pAspNetCoreApplication;
+    ASPNETCORE_CONFIG::GetConfig(pHttpContext, &config);
+
+    // TODO store whether we are inproc or outofproc so we don't need to check the config everytime?
+    if (config->QueryHostingModel()->Equals(L"inprocess", true)) {
```
davidfowl: direct? intgrated pipeline 😄

src/AspNetCore/Inc/fx_ver.h
Eilon: This file is ok. Thanks for the heads up. Any other files?

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,373 @@
...
+    // What should the timeout be? (This is sorta hacky)
+    const HANDLE pHandles[2]{ m_hThread, m_InitalizeEvent };
+
+    // Wait on either the thread to complete or the event to be set
+    dwResult = WaitForMultipleObjects(2, pHandles, FALSE, dwTimeout);
+
+    // It all timed out
+    if (dwResult == WAIT_TIMEOUT)
+    {
+        return E_FAIL;
```
pan-wang: pleas return HRESULT_FROM_WIN32(dwResult) instead of general E_FAIL. It will help diagnose the failure

src/AspNetCore/Inc/fx_ver.h
jkotalik: The associated C file is also copied. Also the logic for finding the hostfxr is copied, but will be translated to the C style used in ANCM.

src/AspNetCore/Src/proxymodule.cxx
```diff
@@ -79,32 +79,82 @@ CProxyModule::OnExecuteRequestHandler(
...
+
+        // TODO: Optimize sync completions
+        return pAspNetCoreApplication->ExecuteRequest(pHttpContext);
+    Failed:
+        pHttpContext->GetResponse()->SetStatus(500, "Internal Server Error", 0, E_APPLICATION_ACTIVATION_EXEC_FAILURE);
         return RQ_NOTIFICATION_FINISH_REQUEST;
     }
-
-    return m_pHandler->OnExecuteRequestHandler();
+    else
```
pan-wang: this is too general. user may give random string now and later found behavior changed.  It's better check empty or specific string value with else if and then return error for other scenario

src/AspNetCore/Src/proxymodule.cxx
jkotalik: Huh?

jkotalik:
@pan-wang I updated aspnetcoreapplication.cxx and .h to the correct style (mostly); feel free to take a look.

AspNetCoreModule.sln
jkotalik: Mistake.

src/AspNetCore/Inc/fx_ver.h
Eilon: fx_ver.c is fine too. What's the other file you're talking about? What's the original file it came from?

src/AspNetCore/Inc/application.h
```diff
@@ -181,6 +183,51 @@ class APPLICATION
...
+
+        if (m_pAspNetCoreApplication == NULL)
+        {
+            AcquireSRWLockExclusive(&m_srwLock);
+            fLockTaken = TRUE;
+
+            if (m_pAspNetCoreApplication == NULL)
+            {
+                application = new ASPNETCORE_APPLICATION();
+                if (application == NULL) {
```
pan-wang, line 206: you failed to create a new application. It should be an error while you still return S_OK as you did not set the hr value to out of memory

src/AspNetCore/Inc/fx_ver.h
jkotalik: The logic for finding the [highest dotnet version](https://github.com/aspnet/AspNetCoreModule/pull/152/files#diff-97748fd8ba0e3e9583d9ff0e1cb6ea2aR459) was originally taken from [here](https://github.com/dotnet/core-setup/blob/83ca8e67d3e830e3e689c722468c620b2a99ba81/src/corehost/corehost.cpp#L104).

Eilon: Ok that's legit too. Thanks for checking!

src/AspNetCore/Src/aspnetcoreconfig.cxx
```diff
@@ -144,13 +150,19 @@ ASPNETCORE_CONFIG::Populate(
...
 
     pAdminManager = g_pHttpServer->GetAdminManager();
-
     hr = strSiteConfigPath.Copy(pHttpContext->GetApplication()->GetAppConfigPath());
     if (FAILED(hr))
     {
         goto Finished;
     }
 
+    BSTR test = SysAllocString(pHttpContext->GetApplication()->GetApplicationPhysicalPath());
```
pan-wang: why do you need this BSTR test?

jkotalik: Poor naming and most likely this should be a STRU. We need the full application path to run it with dotnet.exe.

src/AspNetCore/Src/proxymodule.cxx
```diff
@@ -79,32 +78,87 @@ CProxyModule::OnExecuteRequestHandler(
...
+        // Disable response buffering by default, we'll do a write behind buffering in managed code
+        ((IHttpResponse2*)pHttpContext->GetResponse())->DisableBuffering();
+
+        // TODO: Optimize sync completions
+        return pAspNetCoreApplication->ExecuteRequest(pHttpContext);
+    Failed:
+        pHttpContext->GetResponse()->SetStatus(500, "Internal Server Error", 0, E_APPLICATION_ACTIVATION_EXEC_FAILURE);
+        return REQUEST_NOTIFICATION_STATUS::RQ_NOTIFICATION_FINISH_REQUEST;
+    }
+    return REQUEST_NOTIFICATION_STATUS::RQ_NOTIFICATION_FINISH_REQUEST;
```
pan-wang: do you want to return empty 200 response or 500 in case user gave unknown hostingModel value? it's wired to return 200 now. I think you should do goto Failed with some hr indicating configuration error

src/AspNetCore/Src/proxymodule.cxx
```diff
@@ -79,32 +78,87 @@ CProxyModule::OnExecuteRequestHandler(
...
+
+    // TODO store whether we are inproc or outofproc so we don't need to check the config everytime?
+    if (config->QueryHostingModel()->IsEmpty() || config->QueryHostingModel()->Equals(L"outofprocess", TRUE))// case insensitive
     {
-        pHttpContext->GetResponse()->SetStatus(500, "Internal Server Error", 0, E_OUTOFMEMORY);
-        return RQ_NOTIFICATION_FINISH_REQUEST;
+        m_pHandler = new FORWARDING_HANDLER(pHttpContext);
+        if (m_pHandler == NULL)
+        {
+            pHttpContext->GetResponse()->SetStatus(500, "Internal Server Error", 0, E_OUTOFMEMORY);
```
pan-wang: best practice is to "goto Failed" as you have duplicated code t line 132. you can set the hr = E_OUTOFMEMORY

src/AspNetCore/Src/proxymodule.cxx
```diff
@@ -79,32 +78,87 @@ CProxyModule::OnExecuteRequestHandler(
...
-    return m_pHandler->OnAsyncCompletion(
+    if (config->QueryHostingModel()->IsEmpty() || config->QueryHostingModel()->Equals(L"outofprocess", TRUE))// case insensitive
+    {
+        return m_pHandler->OnAsyncCompletion(
             pCompletionInfo->GetCompletionBytes(),
             pCompletionInfo->GetCompletionStatus());
+    }
+    else if (config->QueryHostingModel()->Equals(L"inprocess", TRUE)) // case insensitive
+    {
+        return REQUEST_NOTIFICATION_STATUS::RQ_NOTIFICATION_CONTINUE;
```
pan-wang: do you always want to return continue even for failure case, e.g., 500 was set? Best practice is to return Finished_Request in case of error to shortcut IIS pipeline.

src/AspNetCore/Inc/aspnetcoreapplication.h
```diff
@@ -0,0 +1,112 @@
...
+    // The shutdown handler callback from managed code
+    PFN_SHUTDOWN_HANDLER m_ShutdownHandler;
+    void* m_ShutdownHandlerContext;
+    // The event that gets triggered when managed initialization is complete
+    HANDLE m_InitalizeEvent;
+
+    // The exit code of the .NET Core process
+    int m_ProcessExitCode;
+
+    static ASPNETCORE_APPLICATION* s_Application;
```
pan-wang: have you try hosting two in-proc application under same site? this static instance may cause problem

src/AspNetCore/Inc/fx_ver.h
Eilon: Ok I'm unsubscribing from this thread to avoid spam, but @ mention me again if you need me to check something.

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,516 @@
...
+    _Out_ DWORD* cbBytes,
+    _Out_ HRESULT* hr
+)
+{
+    *cbBytes = info->GetCompletionBytes();
+    *hr = info->GetCompletionStatus();
+}
+
+extern "C" __declspec(dllexport)
+BSTR // TODO probably should make this a wide string
```
pan-wang: why do you want to return BSTR a copy of the memory? Is it required by managed layer? or some tried to change the value?

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,516 @@
...
+http_write_response_bytes(
+    _In_ IHttpContext* pHttpContext,
+    _In_ HTTP_DATA_CHUNK* pDataChunks,
+    _In_ DWORD nChunks,
+    _In_ PFN_ASYNC_COMPLETION pfnCompletionCallback,
+    _In_ VOID* pvCompletionContext,
+    _In_ BOOL* pfCompletionExpected
+)
+{
+    auto pHttpResponse = (IHttpResponse2*)pHttpContext->GetResponse();
```
pan-wang: please avoid auto

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,516 @@
...
+
+    BOOL fAsync = TRUE;
+    BOOL fMoreData = TRUE;
```
pan-wang: why do you hardcode those values

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,516 @@
...
+    else
+    {
+        dwTimeout = pConfig->QueryStartupTimeLimitInMS();
+    }
+
+    // What should the timeout be? (This is sorta hacky)
+    const HANDLE pHandles[2]{ m_hThread, m_InitalizeEvent };
+
+    // Wait on either the thread to complete or the event to be set
+    dwResult = WaitForMultipleObjects(2, pHandles, FALSE, dwTimeout);
```
pan-wang: this block of code looks suspicious. if WaitForMultipleObjects returns WAIT_FAILED or WAIT_ABANDONED_0, the code logic seems wrong

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,516 @@
...
+}
+
+// Thread execution callback
+static
+VOID
+ExecuteAspNetCoreProcess(
+    _In_ LPVOID pContext
+)
+{
+    auto pApplication = (ASPNETCORE_APPLICATION*)pContext;
```
pan-wang: please don't use auto. We know which type is being used.

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,516 @@
...
+    PCWSTR          pszDotnetExeLocation = NULL;
+    PCWSTR          pszDotnetExeString(L"dotnet.exe");
+    DWORD           dwCopyLength;
+    HMODULE         hModule;
+    PCWSTR          argv[2];
+    hostfxr_main_fn pProc;
+
+    // Get the System PATH value.
+    hr = GetEnv(L"PATH", &strFullPath);
+    if (FAILED(hr))
```
pan-wang: what would be the http response if ExecuteApplication failed in the middle? I am really worry about it

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,516 @@
...
+        goto Failed;
+    }
+
+    pszBuffer = new WCHAR[dwLength];
+    if (GetEnvironmentVariableW(pszEnvironmentVariable, pszBuffer, dwLength) == 0)
+    {
+        goto Failed;
+    }
+
+    pstrResult->Copy(pszBuffer);
```
pan-wang: this return will cause memory leak as pszBuffer was not freed

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,516 @@
...
+        // which doesn't need the dotnet.exe portion of the string
+        // TODO consider reducing allocations.
+        strDotnetExeLocation.Reset();
+        strDotnetFolderLocation.Reset();
+        strDotnetExeLocation.Copy(pszDotnetExeLocation, dwCopyLength);
+        strDotnetFolderLocation.Copy(pszDotnetExeLocation, dwCopyLength);
+
+        if (dwCopyLength > 0 && pszDotnetExeLocation[dwCopyLength - 1] != L'\\')
+        {
+            strDotnetExeLocation.Append(L"\\");
```
pan-wang: please check the returned hresult

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,516 @@
...
+        }
+
+        strDotnetExeLocation.Append(pszDotnetExeString);
```
pan-wang: please check the returned hr of append

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,516 @@
...
+        if (PathFileExists(strDotnetExeLocation.QueryStr()))
+        {
+            break;
+        }
+        pszDotnetExeLocation = wcstok_s(NULL, L";", &strDelimeterContext);
+    }
+    strDotnetFolderLocation.Append(L"\\host\\fxr");
```
pan-wang: missing hresult check here

src/AspNetCore/Inc/aspnetcoreapplication.h
jkotalik: For sure. I tried the two in-proc applications and the second app just returns a 500 server error (going to investigate further tomorrow).

src/AspNetCore/Src/proxymodule.cxx
jkotalik: @davidfowl  we probably need to add an OnAsyncCompletion handler to the AspNetCoreApplication?

src/AspNetCore/Src/aspnetcoreapplication.cxx
jkotalik: The value is required for the managed layer to know where the content root is. Before I had this returned as an out parameter, however I was running into Heap Corruption (I think both managed and native were freeing). @davidfowl did you manage to follow up with whoever is knowledgeable about this?

src/AspNetCore/Src/aspnetcoreapplication.cxx
jkotalik: @davidfowl Can we just assume that async and more data will always be TRUE?

src/AspNetCore/Src/aspnetcoreapplication.cxx
jkotalik: 500 server error?

src/AspNetCore/Src/aspnetcoreconfig.cxx
```diff
@@ -8,11 +8,16 @@ ASPNETCORE_CONFIG::~ASPNETCORE_CONFIG()
     //
     // the destructor will be called once IIS decides to recycle the module context (i.e., application)
     //
+    // shutting down core application first
+    if (ASPNETCORE_APPLICATION::GetInstance() != NULL) {
```
pan-wang, line 12: this will cause problem in case of more than one applications

src/AspNetCore/Src/proxymodule.cxx
```diff
@@ -79,32 +78,87 @@ CProxyModule::OnExecuteRequestHandler(
...
+        ((IHttpContext3*)pHttpContext)->EnableFullDuplex();
+
+        // Disable response buffering by default, we'll do a write behind buffering in managed code
+        ((IHttpResponse2*)pHttpContext->GetResponse())->DisableBuffering();
+
+        // TODO: Optimize sync completions
+        return pAspNetCoreApplication->ExecuteRequest(pHttpContext);
+    }
+Failed:
+    pHttpContext->GetResponse()->SetStatus(500, "Internal Server Error", 0, E_APPLICATION_ACTIVATION_EXEC_FAILURE);
```
pan-wang: its better to use hr instead of hardcoded Exec_Failure

src/AspNetCore/Src/aspnetcoreapplication.cxx
davidfowl: The intent is to wait on any of the events to fire

src/AspNetCore/Src/aspnetcoreconfig.cxx
davidfowl: We discussed this right? It needs to throw if there's different apps in the same app pool.

pan-wang: so our hosting mode is one application per apppool? it's hard to force this rule and outproc does support multiple applications inside same apppool

davidfowl: That has been the recommended model for years. Guidance for hosters is one application per application pool. In this particular case, we need to detect more than one and throw because we don't have or want to host with app domains.

pan-wang: if user somehow hosts two applications, it then totally depends which one app receives the request first. The behavior is unpredictable. We may receive a lot of customer reports.  It will be a MUST to log a specific error to windows log for the failure case.

pan-wang: I understand supporting multiple apps is not a goal for this release. Have we try it without appdomain and know where the blocker is?

src/AspNetCore/Src/aspnetcoreapplication.cxx
jkotalik: Why did we have the InitializeEvent to begin with? To detect if we have already initialized the AspNetCoreApplication?

src/AspNetCore/Src/proxymodule.cxx
```diff
@@ -79,32 +78,87 @@ CProxyModule::OnExecuteRequestHandler(
...
+        }
+
+        hr = pApplicationManager->GetApplication(pHttpContext,
+            &pApplication);
+        if (FAILED(hr))
+        {
+            goto Failed;
+        }
+
+
```
jkotalik: Add a static counter such that we don't try to create the application more than ~5 times.

src/AspNetCore/Inc/aspnetcoreapplication.h
```diff
@@ -0,0 +1,117 @@
...
+
+    ASPNETCORE_APPLICATION():
+        m_pConfiguration(NULL),
+        m_RequestHandler(NULL)
+    {
+    }
+
+    ~ASPNETCORE_APPLICATION()
+    {
+    }
```
pan-wang: please make sure all handles get closed here to avoid handle leak and all resources get freed

src/AspNetCore/Src/aspnetcoreapplication.cxx
pan-wang: you can always assume async. However, you need to define the behavior of empty chunk and null chunk, for empty chunk, you may want to set fmoredata as false.

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,590 @@
...
+    {
+        goto Failed;
+    }
+
+    // Split on ';', checking to see if dotnet.exe exists in any folders.
+    pszDotnetExeLocation = wcstok_s(strFullPath.QueryStr(), L";", &strDelimeterContext);
+
+    while (pszDotnetExeLocation != NULL)
+    {
+        dwCopyLength = wcsnlen_s(pszDotnetExeLocation, 260);
```
pan-wang, line 327: this will break the longpath scenario. You can use strFullPath.QueryCCH()

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,590 @@
...
+    hModule = LoadLibraryW(strDotnetFolderLocation.QueryStr());
+
+    if (hModule == nullptr)
+    {
+        // .NET Core not installed (we can log a more detailed error message here)
+        goto Failed;
+    }
+
+    // Get the entry point for main
+    pProc = (hostfxr_main_fn)GetProcAddress(hModule, "hostfxr_main");
```
pan-wang, line 439: null check for pProc

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,590 @@
...
+    // until this initialization is done.
+
+    // We set a static so that managed code can call back into this instance and
+    // set the callbacks
+    s_Application = this;
+
+    m_ProcessExitCode = pProc(2, argv);
+    if (m_ProcessExitCode != 0)
+    {
+        // TODO error
```
pan-wang: please open an issue to track the work

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,590 @@
...
+    // set the callbacks
+    s_Application = this;
+
+    m_ProcessExitCode = pProc(2, argv);
+    if (m_ProcessExitCode != 0)
+    {
+        // TODO error 
+    }
+
+    return hr;
```
pan-wang: no need for this return here as failed will be executed. It's better to rename Failed to Finished as you do not handle failure

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,590 @@
...
+    if (pszEnvironmentVariable == NULL)
+    {
+        return FALSE;
+    }
+    pstrResult->Reset();
+    dwLength = GetEnvironmentVariableW(pszEnvironmentVariable, NULL, 0);
+
+    if (dwLength == 0)
+    {
+        if (GetLastError() != ERROR_ENVVAR_NOT_FOUND)
```
pan-wang: this means the envvar was defined but value is empty. Not sure what you plan to do with GetLastError as no_op for current code. Please add some comments for code maintenance

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,590 @@
...
+
+    pszBuffer = new WCHAR[dwLength];
+    if (GetEnvironmentVariableW(pszEnvironmentVariable, pszBuffer, dwLength) == 0)
+    {
+        goto Failed;
+    }
+
+    pstrResult->Copy(pszBuffer);
+
+    if (pszBuffer != NULL) {
```
pan-wang: you can add a local bool var to hold the check result and return it from Failed. This will save the duplicated code and have better code readability

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,590 @@
...
+    fx_ver_t max_ver(-1, -1, -1);
+    for (const auto& dir : vFolders)
+    {
+        fx_ver_t fx_ver(-1, -1, -1);
+        if (fx_ver_t::parse(dir, &fx_ver, false))
+        {
+            max_ver = std::max(max_ver, fx_ver);
+        }
+    }
+    pstrResult->Copy(max_ver.as_str().c_str());
```
pan-wang: You need to handle copy failure. Maybe this method should return hresult

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,590 @@
...
+}
+
+
+BOOL
+ASPNETCORE_APPLICATION::DirectoryExists(
+    _In_ STRU *pstrPath
+)
+{
```
pan-wang, line 566: what happen if the input value is . or ..

src/AspNetCore/Inc/aspnetcoreconfig.h
```diff
@@ -166,6 +175,19 @@ class ASPNETCORE_CONFIG : IHttpStoredContext
...
+    BOOL
+    QueryIsInProcess()
+    {
+        return m_fIsInProcess;
+    }
+
+    BOOL
+    QueryIsOutOfProcess()
+    {
+        return m_fIsOutOfProcess;
```
pan-wang, line 187: as your code set this value based on configuration string, it's better to initialize them at the constructor

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,590 @@
...
+    WIN32_FILE_ATTRIBUTE_DATA data;
+
+    if (pstrPath->QuerySizeCCH() == 0)
```
pan-wang: you should call IsEmpty()  QuerySizeCCH return the buffer size in char

src/AspNetCore/Src/aspnetcoreapplication.cxx
```diff
@@ -0,0 +1,596 @@
...
+
+
+VOID
+ASPNETCORE_APPLICATION::Shutdown(
+    VOID
+)
+{
+    // First call into the managed server and shutdown
+    BOOL result = m_ShutdownHandler(m_ShutdownHandlerContext);
+
```
pan-wang: you need set the static instance to null in order to avoid AV

jkotalik:
@pan-wang I updated this based on your feedback.

src/AspNetCore/Src/aspnetcoreapplication.cxx
jkotalik: strFullPath contains all path variables. Should I wrap pszDotnetExeLocation in a stru?

src/AspNetCore/Src/aspnetcoreapplication.cxx
jkotalik: Most of these methods should probably be private static?

src/AspNetCore/Inc/aspnetcoreconfig.h
jkotalik: Are you referring to the private constructor or the Populate function?

src/AspNetCore/Src/aspnetcoreconfig.cxx
jkotalik: > Have we try it without appdomain and know where the blocker is?

I mean either way the app will fail to load as the static ASPNETCORE_APPLICATION is already set.

jkotalik:
@pan-wang Let's talk tomorrow about how to handle logging errors. I'll go ahead and file an issue, however that feature is of higher priority.

src/AspNetCore/Src/aspnetcoreconfig.cxx
davidfowl: @pan-wang There are no appdomains on .NET Core.

pan-wang (approved):
let's check in now and then I will do the refactoring

PR #153, opened 2017-09-12
Added new test scenarios for websocket

jhkimnew:
@pan-wang Please approve this for the new test scenarios for websocket scenarios and fixing build issues.


test/AspNetCoreModule.Test/Framework/IISConfigUtility.cs
```diff
@@ -254,7 +271,7 @@ public void RecycleAppPool(string appPoolName)
             using (ServerManager serverManager = GetServerManager())
             {
                 serverManager.ApplicationPools[appPoolName].Recycle();
-            }
+            }
```
pan-wang: Remove the whitespaces

test/AspNetCoreModule.Test/Framework/TestUtility.cs
```diff
@@ -811,7 +851,7 @@ public static int RunCommand(string fileName, string arguments = null, bool chec
             p.StartInfo.UseShellExecute = false;
             p.StartInfo.CreateNoWindow = true;            
             p.Start(); 
-            pid = p.Id;
+            pid = p.Id;
```
pan-wang: Remove the whitespaces

test/AspNetCoreModule.Test/Framework/TestWebSite.cs
```diff
@@ -309,41 +353,126 @@ public void StartIISExpress(string verificationCommand = null)
...
                 {
-                    isIISExpressReady = true;
-                    break;
+                    throw new System.ApplicationException("Not found :" + debuggerCmdline + "; this test requires windbg.exe.");
                 }
-                else
+            }
+            else
+            {
+                debuggerCmdline = Path.Combine(Environment.ExpandEnvironmentVariables("%ProgramFiles%"), "Debugging Tools for Windows (x64)", "windbg.exe");
```
pan-wang, line 414: you hardcode the path of WinDBG with default installation path. As user may change the installation path, you may want to  run query such as https://stackoverflow.com/questions/908850/get-installed-applications-in-a-system

pan-wang (approved):

test/AspNetCoreModule.Test/Framework/TestWebSite.cs
jhkimnew: I have a plan to update whole tests can be started with attaching debugger and appverifer rule. At that time, I will remove the hardcode path as you suggest.

PR #172, opened 2017-09-27
ANCM Test Refactoring: Initial check-in to fix the build issue on AspNetCI

jhkimnew:
@pan-wang @jkotalik 
This is the initial test code fix for ASPNet CI. 
Experimentally, I set only two test cases with P0 so that I can verify those tests got passed in CI build machine. I will set P0 for other test cases later.

### DIFF
--- a/test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
+++ b/test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
@@ -5,34 +5,52 @@ using System;
 using System.IO;
 using System.Threading;
 using Microsoft.Extensions.PlatformAbstractions;
+using System.Security.Principal;
+using System.Security.AccessControl;
 
 namespace AspNetCoreModule.Test.Framework
 {
+    public static class ANCMTestFlags
+    {
+        public const string TestSkipContext = "SkipTest";
+        public const string UsePrivateANCM = "UsePrivateANCM";
+        public const string UseIISExpressContext = "UseIISExpress";
+        public const string UseFullIISContext = "UseFullIIS";
+        public const string RunAsNonAdminUser = "RunAsNonAdminUser";
+        public const string MakeCertExeAvailable = "MakeCertExeAvailable";
+        public const string X86Platform = "X86Platform";
+        public const string Wow64BitMode = "Wow64BitMode";
+        public const string P0 = "P0";
+        public const string P1 = "P1";
+    }
+
     public class InitializeTestMachine : IDisposable
     {
         public const string ANCMTestFlagsEnvironmentVariable = "%ANCMTestFlags%";
-        public const string ANCMTestFlagsDefaultContext = "AdminAnd64Bit";
-        public const string ANCMTestFlagsTestSkipContext = "SkipTest";
-        public const string ANCMTestFlagsUsePrivateAspNetCoreFileContext = "UsePrivate";
-        private const string ANCMTestFlagsUseIISExpressContext = "UseIISExpress";
+
+        public const string ANCMTestFlags_DefaultContext_IISExpressP0Only = ANCMTestFlags.UsePrivateANCM
+                                                        + ";"
+                                                        + ANCMTestFlags.UseIISExpressContext
+                                                        + ";"
+                                                        + ANCMTestFlags.RunAsNonAdminUser
+                                                        + ";"
+                                                        + ANCMTestFlags.P0;
+
+        public const string ANCMTestFlags_DefaultContext_FullIIS = ANCMTestFlags.UsePrivateANCM
+                                                        + ";"
+                                                        + ANCMTestFlags.P0
+                                                        + ";"
+                                                        + ANCMTestFlags.P1;
 
         private static bool? _usePrivateAspNetCoreFile = null;
         public static bool? UsePrivateAspNetCoreFile
         {
             get {
-                // 
-                // By default, we don't use the private AspNetCore.dll that is compiled with this solution.
-                // In order to use the private file, you should add 'UsePrivateAspNetCoreFile' flag to the Environmnet variable %ANCMTestFlag%.
-                //
-                //     Set ANCMTestFlag=%ANCMTestFlag%;UsePrivateAspNetCoreFile 
-                //     Or
-                //     $Env:ANCMTestFlag=$Env:ANCMTestFlag + ";UsePrivateAspNetCoreFile"
-                //
                 if (_usePrivateAspNetCoreFile == null)
                 {
                     _usePrivateAspNetCoreFile = false;
-                    var envValue = Environment.ExpandEnvironmentVariables(ANCMTestFlagsEnvironmentVariable);
-                    if (envValue.ToLower().Contains(ANCMTestFlagsUsePrivateAspNetCoreFileContext.ToLower()))
+                    var envValue = GlobalTestFlags;
+                    if (envValue.ToLower().Contains(ANCMTestFlags.UsePrivateANCM.ToLower()))
                     {
                         TestUtility.LogInformation("PrivateAspNetCoreFile is set");
                         _usePrivateAspNetCoreFile = true;
@@ -55,8 +73,8 @@ namespace AspNetCoreModule.Test.Framework
         public static string FullIisAspnetcore_path = Path.Combine(Environment.ExpandEnvironmentVariables("%windir%"), "system32", "inetsrv", PrivateFileName);
         public static string FullIisAspnetcore_path_original = Path.Combine(Environment.ExpandEnvironmentVariables("%windir%"), "system32", "inetsrv", "aspnetcore.dll");
         public static string FullIisAspnetcore_X86_path = Path.Combine(Environment.ExpandEnvironmentVariables("%windir%"), "syswow64", "inetsrv", PrivateFileName);
-        public static string IisExpressAspnetcore_path = Path.Combine(Environment.ExpandEnvironmentVariables("%ProgramFiles%"), "IIS Express", PrivateFileName);
-        public static string IisExpressAspnetcore_X86_path = Path.Combine(Environment.ExpandEnvironmentVariables("%ProgramFiles(x86)%"), "IIS Express", PrivateFileName);
+        public static string IisExpressAspnetcore_path;
+        public static string IisExpressAspnetcore_X86_path;
 
         public static string IisExpressAspnetcoreSchema_path = Path.Combine(Environment.ExpandEnvironmentVariables("%ProgramFiles%"), "IIS Express", "config", "schema", "aspnetcore_schema.xml");
         public static string IisExpressAspnetcoreSchema_X86_path = Path.Combine(Environment.ExpandEnvironmentVariables("%ProgramFiles(x86)%"), "IIS Express", "config", "schema", "aspnetcore_schema.xml");
@@ -65,27 +83,135 @@ namespace AspNetCoreModule.Test.Framework
         private static bool _InitializeTestMachineCompleted = false;
         private string _setupScriptPath = null;
         
-        private bool CheckPerquisiteForANCMTest()
+        private static bool? _MakeCertExeAvailable = null;
+        public static bool? MakeCertExeAvailable
         {
-            bool result = true;
-            TestUtility.LogInformation("CheckPerquisiteForANCMTest(): Environment.Is64BitOperatingSystem: {0}, Environment.Is64BitProcess {1}", Environment.Is64BitOperatingSystem, Environment.Is64BitProcess);
-            TestUtility.LogInformation("%ANCMTestFlags%: {0}", Environment.ExpandEnvironmentVariables(ANCMTestFlagsEnvironmentVariable));
-
-            if (Environment.Is64BitOperatingSystem && !Environment.Is64BitProcess)
+            get
             {
-                TestUtility.LogInformation("CheckPerquisiteForANCMTest() Failed: ANCM test should be started with x64 process mode on 64 bit machine; if you run this test on Visual Studio, you should set X64 first after selecting 'Test -> Test Settings -> Default Process Architecture' menu");
-                result = false;
+                if (_MakeCertExeAvailable == null)
+                {
+                    _MakeCertExeAvailable = false;
+                    try
+                    {
+                        string makecertExeFilePath = TestUtility.GetMakeCertPath();
+                        TestUtility.RunCommand(makecertExeFilePath, null, true, true);
+                        TestUtility.LogInformation("Verified makecert.exe is available : " + makecertExeFilePath);
+                        _MakeCertExeAvailable = true;
+                    }
+                    catch
+                    {
+                        _MakeCertExeAvailable = false;
+                    }
+                }
+                return _MakeCertExeAvailable;
             }
-            return result;
         }
+
+        public static string TestRootDirectory
+        {
+            get
+            {
+                return Path.Combine(Environment.ExpandEnvironmentVariables("%SystemDrive%") + @"\", "_ANCMTest");
+            }
+        }
+
+        private static string _globalTestFlags = null;
+        public static string GlobalTestFlags
+        {
+            get
+            {
+                if (_globalTestFlags == null)
+                {
+                    bool isElevated;
+                    WindowsIdentity identity = WindowsIdentity.GetCurrent();
+                    WindowsPrincipal principal = new WindowsPrincipal(identity);
+                    isElevated = principal.IsInRole(WindowsBuiltInRole.Administrator);
+
+                    // check if this test process is started with the Run As Administrator start option
+                    _globalTestFlags = Environment.ExpandEnvironmentVariables(ANCMTestFlagsEnvironmentVariable);
+                    _globalTestFlags = _globalTestFlags.ToLower();
+                    if (_globalTestFlags.Contains(ANCMTestFlags.UseIISExpressContext.ToLower()) && _globalTestFlags.Contains(ANCMTestFlags.UseFullIISContext.ToLower()))
+                    {
+                        throw new ApplicationException(ANCMTestFlagsEnvironmentVariable + " can't be set with both "
+                            + ANCMTestFlags.UseIISExpressContext + " and " + ANCMTestFlags.UseFullIISContext);
+                    }
+
+                    // set default context with ANCMTestFlags_DefaultContext_IISExpressP0Only
+                    if (!isElevated
+                        || _globalTestFlags.Replace(ANCMTestFlagsEnvironmentVariable.ToLower(), "").Trim() == ""
+                        || _globalTestFlags.Contains(ANCMTestFlags.UseIISExpressContext.ToLower()))
+                    {
+                        _globalTestFlags = ANCMTestFlags_DefaultContext_IISExpressP0Only;
+                    }
+
+                    // if users set UseFullIIS, the default value should be overwritten with ANCMTestFlags_DefaultContext_FullIIS
+                    if (isElevated
+                        && _globalTestFlags.Contains(ANCMTestFlags.UseFullIISContext.ToLower()))
+                    {
+                        _globalTestFlags = ANCMTestFlags_DefaultContext_FullIIS;
+                    }
+
+                    // adjust the default test context in run time to figure out wrong test context values
+                    if (!isElevated)
+                    {
+                        if (_globalTestFlags.Contains(ANCMTestFlags.UseFullIISContext.ToLower()))
+                        {
+                            _globalTestFlags = _globalTestFlags.Replace(ANCMTestFlags.UseFullIISContext.ToLower(), "");
+                        }
+
+                        if (!_globalTestFlags.Contains(ANCMTestFlags.UseIISExpressContext.ToLower()))
+                        {
+                            TestUtility.LogInformation("Added test context of " + ANCMTestFlags.UseIISExpressContext);
+                            _globalTestFlags += ";" + ANCMTestFlags.UseIISExpressContext;
+                        }
+
+                        if (!_globalTestFlags.Contains(ANCMTestFlags.RunAsNonAdminUser.ToLower()))
+                        {
+                            TestUtility.LogInformation("Added test context of " + ANCMTestFlags.RunAsNonAdminUser);
+                            _globalTestFlags += ";" + ANCMTestFlags.RunAsNonAdminUser;
+                        }
+                    }
+
+                    if (MakeCertExeAvailable == true)
+                    {
+                        if (!_globalTestFlags.Contains(ANCMTestFlags.MakeCertExeAvailable.ToLower()))
+                        {
+                            TestUtility.LogInformation("Added test context of " + ANCMTestFlags.MakeCertExeAvailable);
+                            _globalTestFlags += ";" + ANCMTestFlags.MakeCertExeAvailable;
+                        }
+                    }
+
+                    if (!Environment.Is64BitOperatingSystem)
+                    {
+                        if (!_globalTestFlags.Contains(ANCMTestFlags.X86Platform.ToLower()))
+                        {
+                            TestUtility.LogInformation("Added test context of " + ANCMTestFlags.X86Platform);
+                            _globalTestFlags += ";" + ANCMTestFlags.X86Platform;
+                        }
+                    }
+
+                    if (Environment.Is64BitOperatingSystem && !Environment.Is64BitProcess)
+                    {
+                        if (!_globalTestFlags.Contains(ANCMTestFlags.Wow64BitMode.ToLower()))
+                        {
+                            TestUtility.LogInformation("Added test context of " + ANCMTestFlags.Wow64BitMode);
+                            _globalTestFlags += ";" + ANCMTestFlags.Wow64BitMode;
+                        }
+                    }
+
+                    _globalTestFlags = _globalTestFlags.ToLower();
+                }
+
+                return _globalTestFlags;
+            }            
+        }
+
         public InitializeTestMachine()
         {
             _referenceCount++;
 
             if (_referenceCount == 1)
             {
-                CheckPerquisiteForANCMTest();
-
                 TestUtility.LogInformation("InitializeTestMachine::InitializeTestMachine() Start");
 
                 _InitializeTestMachineCompleted = false;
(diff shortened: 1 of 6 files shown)


jkotalik:
@jhkimnew please sign https://cla2.dotnetfoundation.org/ so the cla-required label goes away 😄

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -5,34 +5,52 @@
...
+    {
+        public const string TestSkipContext = "SkipTest";
+        public const string UsePrivateANCM = "UsePrivateANCM";
+        public const string UseIISExpressContext = "UseIISExpress";
+        public const string UseFullIISContext = "UseFullIIS";
+        public const string RunAsNonAdminUser = "RunAsNonAdminUser";
+        public const string MakeCertExeAvailable = "MakeCertExeAvailable";
+        public const string X86Platform = "X86Platform";
+        public const string Wow64BitMode = "Wow64BitMode";
+        public const string P0 = "P0";
```
jkotalik: P0 and P1 are odd names for privilege levels, @shirhatti can we change this?

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -5,34 +5,52 @@
...
+
+        public const string ANCMTestFlags_DefaultContext_IISExpressP0Only = ANCMTestFlags.UsePrivateANCM
+                                                        + ";"
+                                                        + ANCMTestFlags.UseIISExpressContext
+                                                        + ";"
+                                                        + ANCMTestFlags.RunAsNonAdminUser
+                                                        + ";"
+                                                        + ANCMTestFlags.P0;
+
+        public const string ANCMTestFlags_DefaultContext_FullIIS = ANCMTestFlags.UsePrivateANCM
```
jkotalik: Does the full IIS test require using the private ANCM?

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -5,34 +5,52 @@
...
+                                                        + ";"
+                                                        + ANCMTestFlags.P0
+                                                        + ";"
+                                                        + ANCMTestFlags.P1;
 
         private static bool? _usePrivateAspNetCoreFile = null;
         public static bool? UsePrivateAspNetCoreFile
         {
             get {
```
jkotalik: nit: move open brace to next line.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -65,27 +83,135 @@ public class InitializeTestMachine : IDisposable
         private static bool _InitializeTestMachineCompleted = false;
         private string _setupScriptPath = null;
         
-        private bool CheckPerquisiteForANCMTest()
+        private static bool? _MakeCertExeAvailable = null;
```
jkotalik: nit: lowercase m.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -65,27 +83,135 @@ public class InitializeTestMachine : IDisposable
...
+
+        private static string _globalTestFlags = null;
+        public static string GlobalTestFlags
+        {
+            get
+            {
+                if (_globalTestFlags == null)
+                {
+                    bool isElevated;
+                    WindowsIdentity identity = WindowsIdentity.GetCurrent();
```
jkotalik: These lines can be condense significantly. Ex identity is only used on the next line, so just pass in WindowsIdentity.GetCurrent() into it.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -65,27 +83,135 @@ public class InitializeTestMachine : IDisposable
...
+                    WindowsPrincipal principal = new WindowsPrincipal(identity);
+                    isElevated = principal.IsInRole(WindowsBuiltInRole.Administrator);
+
+                    // check if this test process is started with the Run As Administrator start option
+                    _globalTestFlags = Environment.ExpandEnvironmentVariables(ANCMTestFlagsEnvironmentVariable);
+                    _globalTestFlags = _globalTestFlags.ToLower();
```
jkotalik: This may be for a different PR, but instead of lower casing all of the strings, we can maybe compare some of them ignorecase?

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -65,27 +83,135 @@ public class InitializeTestMachine : IDisposable
...
+                    if (_globalTestFlags.Contains(ANCMTestFlags.UseIISExpressContext.ToLower()) && _globalTestFlags.Contains(ANCMTestFlags.UseFullIISContext.ToLower()))
+                    {
+                        throw new ApplicationException(ANCMTestFlagsEnvironmentVariable + " can't be set with both "
```
jkotalik: Probably not a good idea to be throwing exceptions from properties.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -65,27 +83,135 @@ public class InitializeTestMachine : IDisposable
...
+                    {
+                        _globalTestFlags = ANCMTestFlags_DefaultContext_FullIIS;
+                    }
+
+                    // adjust the default test context in run time to figure out wrong test context values
+                    if (!isElevated)
+                    {
+                        if (_globalTestFlags.Contains(ANCMTestFlags.UseFullIISContext.ToLower()))
+                        {
+                            _globalTestFlags = _globalTestFlags.Replace(ANCMTestFlags.UseFullIISContext.ToLower(), "");
```
jkotalik: I'm not a fan of calling Replace like this. Maybe go back and build up the string based on flags?

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -65,27 +83,135 @@ public class InitializeTestMachine : IDisposable
...
+                        }
+
+                        if (!_globalTestFlags.Contains(ANCMTestFlags.RunAsNonAdminUser.ToLower()))
+                        {
+                            TestUtility.LogInformation("Added test context of " + ANCMTestFlags.RunAsNonAdminUser);
+                            _globalTestFlags += ";" + ANCMTestFlags.RunAsNonAdminUser;
+                        }
+                    }
+
+                    if (MakeCertExeAvailable == true)
```
jkotalik: nit: remove true check.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jkotalik: Maybe make this a method?

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -65,27 +83,135 @@ public class InitializeTestMachine : IDisposable
...
+                    _globalTestFlags = _globalTestFlags.ToLower();
+                }
+
+                return _globalTestFlags;
+            }            
+        }
+
         public InitializeTestMachine()
         {
             _referenceCount++;
```
jkotalik: Can you break up this method into multiple methods? I'm having trouble understanding what it is doing.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -177,10 +296,14 @@ public InitializeTestMachine()
                     TestUtility.LogInformation("We will use IISExpress instead of IIS: " + ex.Message);
                 }
 
-                string siteRootPath = Path.Combine(Environment.ExpandEnvironmentVariables("%SystemDrive%") + @"\", "inetpub", "ANCMTest");
+                string siteRootPath = TestRootDirectory;
                 if (!Directory.Exists(siteRootPath))
                 {
                     Directory.CreateDirectory(siteRootPath);
+                    DirectorySecurity sec = Directory.GetAccessControl(siteRootPath);
```
jkotalik: Will this be hit by running IIS express? If so, it requires an authenticated user, which I think needs Admin permissions?

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -210,7 +333,7 @@ public InitializeTestMachine()
                 
                 if (InitializeTestMachine.UsePrivateAspNetCoreFile == true)
```
jkotalik: nit: remove == true.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -289,7 +412,7 @@ private void RollbackIISApplicationhostConfigFile()
             }
         }
 
-        private void PreparePrivateANCMFiles()
+        private void PreparePrivateANCMFiles(bool isIISReady)
```
jkotalik: This bool is only used once, just read IISConfigUtility.IsIISReady.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -323,30 +446,40 @@ private void PreparePrivateANCMFiles()
...
-                        TestUtility.ResetHelper(ResetHelperMode.StopW3svcStartW3svc);
-                        Thread.Sleep(1000);
+                        //
+                        // NOTE: ANCM schema file can't be overwritten here, if there is any schema change, that should be updated with installing setup
+                        //
 
-                        string from = Path.Combine(outputPath, "x64", "aspnetcore.dll");
-                        TestUtility.FileCopy(from, FullIisAspnetcore_path, overWrite:true, ignoreExceptionWhileDeletingExistingFile:false);
-                        TestUtility.FileCopy(from, IisExpressAspnetcore_path, overWrite: true, ignoreExceptionWhileDeletingExistingFile: false);
+                        if (isIISReady)
```
jkotalik: I'm confused by this check. If IIS is ready, we wait?

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -323,30 +446,40 @@ private void PreparePrivateANCMFiles()
...
-                            from = Path.Combine(outputPath, "Win32", "aspnetcore.dll");
-                            TestUtility.FileCopy(from, FullIisAspnetcore_X86_path, overWrite: true, ignoreExceptionWhileDeletingExistingFile: false);
-                            TestUtility.FileCopy(from, IisExpressAspnetcore_X86_path, overWrite: true, ignoreExceptionWhileDeletingExistingFile: false);
+                            string from_32bit = Path.Combine(outputPath, "Win32", "aspnetcore.dll");
 
-                            // NOTE: schema file can't be overwritten, if there is any schema change, that should be updated manually
-                            from = Path.Combine(outputPath, "Win32", "aspnetcore_schema.xml");
-                            TestUtility.FileCopy(from, IisExpressAspnetcoreSchema_X86_path, overWrite: false, ignoreExceptionWhileDeletingExistingFile: false);
-                        }
+                            if (isIISReady)
```
jkotalik: There are a lot of checks for that bool. Consider separating into two methods rather than checking it multiple times.

test/AspNetCoreModule.Test/Framework/TestUtility.cs
```diff
@@ -263,6 +263,7 @@ public static void LogInformation(string format, params object[] parameters)
         {
             if (format != null)
             {
+                //Logger.LogInformation("[" + DateTime.Now.ToString() + "] " + format, parameters);
```
jkotalik: nit remove this.

test/AspNetCoreModule.Test/Framework/TestWebSite.cs
```diff
@@ -308,10 +327,30 @@ public TestWebSite(IISConfigUtility.AppPoolBitness appPoolBitness, string logger
...
+                            iisConfig.AddModule("AspNetCoreModule", (InitializeTestMachine.IisExpressAspnetcore_path), null);
+                        }
+                    }
+                    else
+                    {
+                        iisConfig.AddModule("AspNetCoreModule", (InitializeTestMachine.IisExpressAspnetcore_path), null);
+                    }
+                }
+
+                string hostname = "";
```
jkotalik: This is repeated twice, abstract to a new method.

test/AspNetCoreModule.Test/FunctionalTestHelper.cs
```diff
@@ -23,64 +23,48 @@
 namespace AspNetCoreModule.Test
 {
     [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
-    public class ANCMTestSkipCondition : Attribute, ITestCondition
+    public class ANCMTestTestPriority : Attribute, ITestCondition
     {
-        private readonly string _environmentVariableName;
-        public ANCMTestSkipCondition(string environmentVariableName)
+        private readonly string _value;
```
jkotalik: nit: what is value here?

test/AspNetCoreModule.Test/FunctionalTestHelper.cs
```diff
@@ -23,64 +23,48 @@
...
-                        if (token.Equals(InitializeTestMachine.ANCMTestFlagsTestSkipContext, StringComparison.InvariantCultureIgnoreCase))
+                        if (token.Equals(ANCMTestFlags.TestSkipContext, StringComparison.InvariantCultureIgnoreCase))
                         {
-                            AdditionalInfo = InitializeTestMachine.ANCMTestFlagsTestSkipContext + " is set";
+                            testSkipFlagFound = true;
                             result = false;
+                            break;
+                        }
+                    }
+                    if (result == false)
```
jkotalik: nit !result

test/AspNetCoreModule.Test/FunctionalTestHelper.cs
```diff
@@ -136,11 +122,12 @@ public enum ReturnValueType
                 {
                     BaseAddress = testSite.AspNetCoreApp.GetUri(),
                     Timeout = TimeSpan.FromSeconds(5),
-                };
+                };
```
jkotalik: nit remove extra space.

jkotalik (changes requested):
- We prefer to use var instead of type name for variable declaration.
- I ran these tests locally and only 6 tests ran. We are changing the functionality such that running in an Administrator terminal doesn't run all tests?

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: Okay, updated 
P0 -> Default
P1 -> RequireRunAsAdministrator

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: There is no restriction. Users can choose whether the private file is used or not for full IIS test. However, if the test is started on non-admin user, test will use IISExpress automatically and in that case there is no other choice.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: fixed

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: fixed

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: fixed

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: I will revisit this issue when other work items are finished.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: This is fixed.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: I agree but I will handle this in the future.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: Fixed

jkotalik: No like just if(MakeCertExeAvailable)

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: Fixed

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: fixed

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: fixed

test/AspNetCoreModule.Test/FunctionalTestHelper.cs
jhkimnew: Fixed

test/AspNetCoreModule.Test/FunctionalTestHelper.cs
jhkimnew: Fixed

test/AspNetCoreModule.Test/FunctionalTestHelper.cs
jhkimnew: Fixed with a specific name.

test/AspNetCoreModule.Test/Framework/TestUtility.cs
jhkimnew: Fixed

test/AspNetCoreModule.Test/Framework/TestWebSite.cs
jhkimnew: Ah, good catch!!! Fixed.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: This is not for waiting. Just double-check if the private file is successfully copied and this is only for Full IIS Server. Code is now updated to make it more clear.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: Fixed.

jhkimnew:
@jkotalik Update is done. Please approve so that I can complete this PR.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: This is already fixed. I think you looked at the old version before I push the new change.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: This does not require Admin permission. I already tested it.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -65,122 +48,260 @@ public class InitializeTestMachine : IDisposable
...
+                    try
+                    {
+                        string makecertExeFilePath = TestUtility.GetMakeCertPath();
+                        TestUtility.RunCommand(makecertExeFilePath, null, true, true);
+                        TestUtility.LogInformation("Verified makecert.exe is available : " + makecertExeFilePath);
+                        _makeCertExeAvailable = true;
+                    }
+                    catch
+                    {
+                        _makeCertExeAvailable = false;
```
jkotalik, line 68: You don't need to set _makeCertExeAvailable to false here.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -65,122 +48,260 @@ public class InitializeTestMachine : IDisposable
...
+                    }
+                }
+                return (_makeCertExeAvailable == true);
             }
-            return result;
```
jkotalik, line 79: Just return _makeCertExeAvailable

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -65,122 +48,260 @@ public class InitializeTestMachine : IDisposable
...
+        {
+            get
+            {
+                if (_globalTestFlags == null)
                 {
-                    System.Diagnostics.Debugger.Launch();                    
-                }
+                    bool isElevated;
+                    WindowsPrincipal principal = new WindowsPrincipal(WindowsIdentity.GetCurrent());
+                    isElevated = principal.IsInRole(WindowsBuiltInRole.Administrator);
```
jkotalik, line 92: No need to declare isElevated before setting it here.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -65,122 +48,260 @@ public class InitializeTestMachine : IDisposable
...
+                isIISInstalled = false;
+            }
+            
+            if (!isIISInstalled)
+            {
+                throw new System.ApplicationException("IIS server is not installed");
+            }
+
+            // Check websocket is installed
+            if (File.Exists(Path.Combine(IISConfigUtility.Strings.IIS64BitPath, "iiswsock.dll")))
```
jkotalik, line 220: The win7 check for websockets occur on each test? This would fail if we were running tests on win7.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -65,122 +48,260 @@ public class InitializeTestMachine : IDisposable
...
+            {
+                TestUtility.LogInformation("Websocket is installed");
+            }
+            else
+            {
+                throw new System.ApplicationException("websocket module is not installed");
```
jkotalik, line 226: Add "using System" to the top of the file and say "throw new ApplicationException("...");" for all of these.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -65,122 +48,260 @@ public class InitializeTestMachine : IDisposable
...
+            {
+                TestUtility.LogInformation("W3SVC service is restarted after restoring applicationhost.config");
+            }
+            else
+            {
+                throw new System.ApplicationException("WWW service can't start");
+            }
+
+            // check URLRewrite module exists
+            if (File.Exists(Path.Combine(IISConfigUtility.Strings.IIS64BitPath, "rewrite.dll")))
```
jkotalik, line 248: For my own knowledge, what does the UrlRewrite module need to be installed for ANCM?

jkotalik: why*

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
```diff
@@ -65,122 +48,260 @@ public class InitializeTestMachine : IDisposable
...
+
+            if (IISConfigUtility.ApppHostTemporaryBackupFileExtention == null)
+            {
+                throw new System.ApplicationException("Failed to backup applicationhost.config");
+            }
+        }
+
+        public InitializeTestMachine()
+        {
+            _referenceCount++;
```
jkotalik, line 265: This is a weird way of doing the singleton pattern. Maybe just have a bool instead or make a new object?

jkotalik (commented):
I'll look more into functional components later.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: Fixed and will send a new commit soon.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: This is by design because the type of MakeCertExeAvalialbe is bool, which is different to _makeCertExeAvailable's (bool?).

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: Fixed and will send a new commit soon.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: Fixed and will send a new commit soon. I made a new test flag of WebSocketModuleAvailable. I will make the websocket test cases requires the flag.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: Fixed and will send a new commit soon.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: Fixed and will send a new commit soon. I made a new test flag of UrlRewriteModuleAvailable. I will make the related test cases to use the flag.
When the urlrewrite rule is enabled, the request sent to IIS is create a new child request based on the UrlRewrite rule. And the child request should be tested to check if it works with ANCM as well.

test/AspNetCoreModule.Test/Framework/InitializeTestMachine.cs
jhkimnew: I will revisit this and remove this. Eventually this will be gone because the singleton logic won't be necessary when the code refactoring is done.

jhkimnew:
@jkotalik Hey, I sent a new commit. Please approve so that I can complete this and check this new change can fix the test run skip issue on AspnetCI agent machine.

jkotalik (approved):

jkotalik:
@jhkimnew I'm working on getting ANCM to run CI tests when you make a pull request rather than just when you push to dev. That way we don't need to verify when merging and rather can check before.

jhkimnew:
+ @pan-wang, @jkotalik  FYI Justin, the build-check on CI when PR is created, not when PR is completed, will be a nice feature for us. Thanks for working on that. 👍

shirhatti:
From a cursory glance, looks like AppVeyor agents have IIS Express installed.
@jkotalik Can we look at enabling these tests on AppVeyor as well?

PR #173, opened 2017-09-29
Disable all tests because of publishing issue on CI agent machine

jhkimnew:
@jkotalik Disable test cases until publishing issue on CI is figured out.


jkotalik:
There is a lot of needless change here for this fix. Let's just do the PR for disabling tests for now.

Issue #181, opened 2017-10-06
Remove IHttpResponse2::WriteEntityChunks as it doesn't guarantee whole response is written on triggering callback.


jkotalik:
When calling WriteEntityChunks with a callback, the callback will be triggered on IOCompletion. If the Dynamic or Static compression modules are on, this callback will be triggered on the first chunk being written, rathern than when the response body is completly written. Also, we cannot reliably chunk up the response body because the dynamic compression chunk size is configurable. As we want to still give the benefits on running behind IIS, we shouldn't tell people to not use the dynamic and static compression modules. So we need to redesign how we write responses, which will involve:
- Storing the managed http context
- Adding a new IHttpStoredContext, containing OnAsyncCompletion, Shutdown, etc.
- Handling websockets (this method should still be safe for websockets, but we may want to simplify the response logic and have everything flow through the same path. 
cc/ @davidfowl @pan-wang

jkotalik:
https://github.com/aspnet/AspNetCoreModule/pull/184

PR #184, opened 2017-10-10
Changes async calls to use OnAsyncComplete event pattern

jkotalik:
Fix for #181. Corresponding PR in IISIntegration: https://github.com/aspnet/IISIntegration/pull/447

### DIFF
--- a/src/AspNetCore/Inc/inprocessapplication.h
+++ b/src/AspNetCore/Inc/inprocessapplication.h
@@ -6,6 +6,7 @@
 typedef void(*request_handler_cb) (int error, IHttpContext* pHttpContext, void* pvCompletionContext);
 typedef REQUEST_NOTIFICATION_STATUS(*PFN_REQUEST_HANDLER) (IHttpContext* pHttpContext, void* pvRequstHandlerContext);
 typedef BOOL(*PFN_SHUTDOWN_HANDLER) (void* pvShutdownHandlerContext);
+typedef REQUEST_NOTIFICATION_STATUS(*PFN_MANAGED_CONTEXT_HANDLER)(void *pvManagedHttpContext, HRESULT hrCompletionStatus, DWORD cbCompletion);
 
 #include "application.h"
 
@@ -39,6 +40,7 @@ public:
     SetCallbackHandles(
         _In_ PFN_REQUEST_HANDLER request_callback,
         _In_ PFN_SHUTDOWN_HANDLER shutdown_callback,
+        _In_ PFN_MANAGED_CONTEXT_HANDLER managed_context_callback,
         _In_ VOID* pvRequstHandlerContext,
         _In_ VOID* pvShutdownHandlerContext
     );
@@ -53,6 +55,13 @@ public:
     LoadManagedApplication(
             VOID
         );
+
+    REQUEST_NOTIFICATION_STATUS
+    OnAsyncCompletion(
+        IHttpContext*           pHttpContext,
+        DWORD                   cbCompletion,
+        HRESULT                 hrCompletionStatus
+    );
 
     static
     IN_PROCESS_APPLICATION*
@@ -76,6 +85,8 @@ private:
     PFN_SHUTDOWN_HANDLER            m_ShutdownHandler;
     VOID*                           m_ShutdownHandlerContext;
 
+    PFN_MANAGED_CONTEXT_HANDLER     m_AsyncCompletionHandler;
+
     // The event that gets triggered when managed initialization is complete
     HANDLE                          m_pInitalizeEvent;
 
@@ -84,6 +95,7 @@ private:
 
     BOOL                            m_fManagedAppLoaded;
     BOOL                            m_fLoadManagedAppError;
+    BOOL                            m_fIsWebSocketsConnection;
 
     static IN_PROCESS_APPLICATION*   s_Application;
 
--- a/src/AspNetCore/Src/forwardinghandler.cxx
+++ b/src/AspNetCore/Src/forwardinghandler.cxx
@@ -1143,11 +1143,6 @@ FORWARDING_HANDLER::OnExecuteRequestHandler(
     {
     case HOSTING_IN_PROCESS:
     {
-        // Allow reading and writing to simultaneously
-        ((IHttpContext3*)m_pW3Context)->EnableFullDuplex();
-
-        // Disable response buffering by default, we'll do a write behind buffering in managed code
-        ((IHttpResponse2*)m_pW3Context->GetResponse())->DisableBuffering();
 
         hr = ((IN_PROCESS_APPLICATION*)m_pApplication)->LoadManagedApplication();
         if (FAILED(hr))
@@ -1570,7 +1565,29 @@ REQUEST_NOTIFICATION_STATUS
             reinterpret_cast<PVOID>(static_cast<DWORD_PTR>(hrCompletionStatus)));
     }
 
-    if (m_pApplication->QueryConfig()->QueryHostingModel() == HOSTING_OUT_PROCESS)
+    // Application is null here? Can this happen for out of proc scenario?
+    if (m_pApplication->QueryConfig()->QueryHostingModel() == HOSTING_IN_PROCESS)
+    {
+        if (FAILED(hrCompletionStatus))
+        {
+            return RQ_NOTIFICATION_FINISH_REQUEST;
+        }
+        else
+        {
+            // For now we are assuming we are in our own self contained box. 
+            // TODO refactor Finished and Failure sections to handle in process and out of process failure.
+            // TODO verify that websocket's OnAsyncCompletion is not calling this.
+            IN_PROCESS_APPLICATION* application = (IN_PROCESS_APPLICATION*)m_pApplication;
+            if (application == NULL)
+            {
+                hr = E_FAIL;
+                return RQ_NOTIFICATION_FINISH_REQUEST;
+            }
+
+            return application->OnAsyncCompletion(m_pW3Context, cbCompletion, hrCompletionStatus);
+        }
+    }
+    else if (m_pApplication->QueryConfig()->QueryHostingModel() == HOSTING_OUT_PROCESS)
     {
         //
         // Take a reference so that object does not go away as a result of
--- a/src/AspNetCore/Src/forwardinghandler.cxx
+++ b/src/AspNetCore/Src/forwardinghandler.cxx
@@ -1565,7 +1565,6 @@ REQUEST_NOTIFICATION_STATUS
             reinterpret_cast<PVOID>(static_cast<DWORD_PTR>(hrCompletionStatus)));
     }
 
-    // Application is null here? Can this happen for out of proc scenario?
     if (m_pApplication->QueryConfig()->QueryHostingModel() == HOSTING_IN_PROCESS)
     {
         if (FAILED(hrCompletionStatus))
--- a/src/AspNetCore/Src/inprocessapplication.cxx
+++ b/src/AspNetCore/Src/inprocessapplication.cxx
@@ -333,9 +333,9 @@ http_cancel_io(
 IN_PROCESS_APPLICATION*  IN_PROCESS_APPLICATION::s_Application = NULL;
 
 IN_PROCESS_APPLICATION::IN_PROCESS_APPLICATION() :
-    m_ProcessExitCode(0),
-    m_fManagedAppLoaded(FALSE),
-    m_fLoadManagedAppError(FALSE)
+    m_ProcessExitCode ( 0 ),
+    m_fManagedAppLoaded ( FALSE ),
+    m_fLoadManagedAppError ( FALSE )
 {
 }
 
--- a/src/AspNetCore/Src/inprocessapplication.cxx
+++ b/src/AspNetCore/Src/inprocessapplication.cxx
@@ -14,6 +14,7 @@ VOID
 register_callbacks(
     _In_ PFN_REQUEST_HANDLER request_handler,
     _In_ PFN_SHUTDOWN_HANDLER shutdown_handler,
+    _In_ PFN_MANAGED_CONTEXT_HANDLER async_completion_handler,
     _In_ VOID* pvRequstHandlerContext,
     _In_ VOID* pvShutdownHandlerContext
 )
@@ -21,6 +22,7 @@ register_callbacks(
     IN_PROCESS_APPLICATION::GetInstance()->SetCallbackHandles(
         request_handler,
         shutdown_handler,
+        async_completion_handler,
         pvRequstHandlerContext,
         pvShutdownHandlerContext
     );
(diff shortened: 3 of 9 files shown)


src/AspNetCore/Inc/forwardinghandler.h
```diff
@@ -70,6 +70,9 @@ class FORWARDING_HANDLER
 
     REQUEST_NOTIFICATION_STATUS
     OnAsyncCompletion(
+        IHttpContext *          pHttpContext,
```
pan-wang: I don't think you need those input parameters. forwardinghandler already has m_pW3Context wrapping the request.
ANCM only listening EXecuteRequestHandler and you will not get other notification.

src/AspNetCore/Inc/inprocessapplication.h
```diff
@@ -54,6 +56,28 @@ class IN_PROCESS_APPLICATION : public APPLICATION
...
+
+
+    HRESULT
+    GetInProcessStoredContext(
+        IHttpContext* pHttpContext,
+        IN_PROCESS_STORED_CONTEXT** ppInProcessStoredContext
+    );
+
+    VOID
+    SetWebsocketsEnabled(
```
pan-wang: you don't need this function. we have a global variable g_fWebSocketSupported which is set to true for win8 and above only

src/AspNetCore/Inc/inprocessapplication.h
```diff
@@ -84,6 +110,7 @@ class IN_PROCESS_APPLICATION : public APPLICATION
 
     BOOL                            m_fManagedAppLoaded;
     BOOL                            m_fLoadManagedAppError;
+    BOOL                            m_fWebSocketsEnabled;
```
pan-wang: no need for it

src/AspNetCore/Src/forwardinghandler.cxx
```diff
@@ -1534,7 +1532,29 @@ REQUEST_NOTIFICATION_STATUS
             reinterpret_cast<PVOID>(static_cast<DWORD_PTR>(hrCompletionStatus)));
     }
 
-    if (m_pApplication->QueryConfig()->QueryHostingModel() == HOSTING_OUT_PROCESS)
+    if (m_pApplication->QueryConfig()->QueryHostingModel() == HOSTING_IN_PROCESS)
+    {
+        IN_PROCESS_APPLICATION* application = (IN_PROCESS_APPLICATION*)m_pApplication;
+        if (application == NULL)
```
pan-wang: you need return finish request here for error confdition

src/AspNetCore/Src/forwardinghandler.cxx
```diff
@@ -1534,7 +1532,29 @@ REQUEST_NOTIFICATION_STATUS
...
+        {
+            hr = E_FAIL; // TODO
+        }
+
+        if (fPostNotification)
```
pan-wang: no need to do post notification check

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -56,10 +58,36 @@ EXTERN_C __MIDL_DECLSPEC_DLLEXPORT VOID http_set_response_status_code(
...
+        return E_OUTOFMEMORY;
+    }
+
+    IHttpModuleContextContainer* pModuleContextContainer = pHttpContext->GetModuleContextContainer();
+    if (pModuleContextContainer == NULL)
+    {
+        return E_FAIL;
+    }
+
+    hr = pModuleContextContainer->SetModuleContext(pInProcessStoredContext, g_pModuleId);
```
pan-wang: if you allow managed code to set this store context more than once, you have to ignore hr == HRESULT_FROM_WIN32(ERROR_ALREADY_ASSIGNED)

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -56,10 +58,36 @@ EXTERN_C __MIDL_DECLSPEC_DLLEXPORT VOID http_set_response_status_code(
...
+)
+{
+    HRESULT hr;
+    IN_PROCESS_STORED_CONTEXT* pInProcessStoredContext = new IN_PROCESS_STORED_CONTEXT(pHttpContext, pvManagedContext);
+    if (pInProcessStoredContext == NULL)
+    {
+        return E_OUTOFMEMORY;
+    }
+
+    IHttpModuleContextContainer* pModuleContextContainer = pHttpContext->GetModuleContextContainer();
```
pan-wang: no need to do null check for this contextcontainer. if it is null, IIS is in broken state.
you can simply do  pHttpContext->GetModuleContextContainer()->SetModuleContext(

src/AspNetCore/Src/inprocessapplication.cxx
jkotalik: It should only set it once.

src/AspNetCore/Inc/inprocessapplication.h
jkotalik: This is state stored on the application to say whether the current stream is websockets or not. If I register a callback to IHttpResponse2::WriteEntityChunks(), will that trigger the OnAsyncCompletion method?

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -107,6 +135,82 @@ http_read_request_bytes(
     _In_ IHttpContext* pHttpContext,
     _In_ CHAR* pvBuffer,
     _In_ DWORD cbBuffer,
+    _In_ DWORD* pDwBytesReceived,
```
pan-wang: lower case dw

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -107,6 +135,82 @@ http_read_request_bytes(
...
+
+    return hr;
+}
+
+EXTERN_C __MIDL_DECLSPEC_DLLEXPORT
+HRESULT
+http_write_response_bytes(
+    _In_ IHttpContext* pHttpContext,
+    _In_ HTTP_DATA_CHUNK* pDataChunks,
+    _In_ DWORD nChunks,
```
pan-wang: dwChunks

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -107,6 +135,82 @@ http_read_request_bytes(
     _In_ IHttpContext* pHttpContext,
     _In_ CHAR* pvBuffer,
     _In_ DWORD cbBuffer,
+    _In_ DWORD* pDwBytesReceived,
+    _In_ BOOL* pfCompletionPending
+)
+{
```
pan-wang: as it is exported method, it's better to check the input parameters, e.g., NULL

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -107,6 +135,82 @@ http_read_request_bytes(
...
+        &dwBytesSent,
+        pfCompletionExpected);
+
+    return hr;
+}
+
+EXTERN_C __MIDL_DECLSPEC_DLLEXPORT
+HRESULT
+http_flush_response_bytes(
+    _In_ IHttpContext* pHttpContext,
```
pan-wang, line 198: should we consider fAsync as input parameter

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -189,13 +293,32 @@ http_flush_response_bytes(
         pfCompletionExpected);
     return hr;
 }
-// End of export 
 
+EXTERN_C __MIDL_DECLSPEC_DLLEXPORT
+HRESULT
+http_enable_websockets(
```
pan-wang, line 307: I think we don't need this function. We can always enable duplex in Forwarding_Handler::OnExecuteRequestHandler for Win8 and above

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -205,6 +328,46 @@ IN_PROCESS_APPLICATION::~IN_PROCESS_APPLICATION()
...
+    DWORD                   cbCompletion,
+    HRESULT                 hrCompletionStatus
+)
+{
+    HRESULT hr;
+    IN_PROCESS_STORED_CONTEXT* pInProcessStoredContext = NULL;
+
+    hr = GetInProcessStoredContext(pHttpContext, &pInProcessStoredContext);
+    if (FAILED(hr))
+    {
```
pan-wang: We need set http 500 error status before calling finish request

src/AspNetCore/Src/inprocessapplication.cxx
jkotalik: WIth the OnAsyncCompletion, we can't have concurrent reads and writes going on at the same time right? Hence only enabling duplex mode when we enable websockets.

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -205,6 +328,46 @@ IN_PROCESS_APPLICATION::~IN_PROCESS_APPLICATION()
     Recycle();
 }
 
+REQUEST_NOTIFICATION_STATUS
+IN_PROCESS_APPLICATION::OnAsyncCompletion(
```
pan-wang, line 348: It does not make sense to me to have this member method. You can directly call it in FORWARDING_HANDLER::OnAsyncCompletion

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -205,6 +328,46 @@ IN_PROCESS_APPLICATION::~IN_PROCESS_APPLICATION()
...
+    {
+        // Finish the request as we couldn't get the callback
+        return RQ_NOTIFICATION_FINISH_REQUEST;
+    }
+    // Call the managed handler for async completion.
+    return m_ManagedContextHandler(pInProcessStoredContext->QueryManagedHttpContext(), hrCompletionStatus, cbCompletion);
+}
+
+HRESULT
+IN_PROCESS_APPLICATION::GetInProcessStoredContext(
```
pan-wang: This method has no dependency on any other members of IN_PROCESS_APPLICATION. You can make it as a standalone method or exported on

src/AspNetCore/Src/proxymodule.cxx
```diff
@@ -130,6 +130,9 @@ CProxyModule::OnAsyncCompletion(
 )
 {
     return m_pHandler->OnAsyncCompletion(
+        pHttpContext,
```
pan-wang: the handler already wrapped the httpcontext. So far we don't need notification type info and fPostNotifcation

src/AspNetCore/Src/inprocessapplication.cxx
pan-wang: On win8 and above, you can process the request and write response with this duplex mode like what you did. Only on Win7, you cannot use it.  This is the internal implementation details on how you handle the request

src/AspNetCore/Src/forwardinghandler.cxx
```diff
@@ -1534,7 +1532,38 @@ REQUEST_NOTIFICATION_STATUS
...
 
-    if (m_pApplication->QueryConfig()->QueryHostingModel() == HOSTING_OUT_PROCESS)
+    if (m_pApplication->QueryConfig()->QueryHostingModel() == HOSTING_IN_PROCESS)
+    {
+        if (fPostNotification)
+        {
+            return RQ_NOTIFICATION_FINISH_REQUEST;
+        }
+
+        if (FAILED(hrCompletionStatus))
```
davidfowl: We shouldn't do this here. What if there's a pending async operation in managed code? It'll never complete.

src/AspNetCore/Src/inprocessapplication.cxx
jkotalik: I'll add this check for now. Should we add it everywhere?

src/AspNetCore/Src/inprocessapplication.cxx
jkotalik: Let's do that in a future PR.

src/AspNetCore/Src/inprocessapplication.cxx
jkotalik: The handler is unique to the application though, so either we need to expose it to the forwarding handler or do something else.

davidfowl:
I'd like to see a test with concurrent reads and writes without websockets. There is a deadlock here.

src/AspNetCore/Src/forwardinghandler.cxx
jkotalik: I think we need to start checking the number of bytes written. We need to also confirm that the result of the read/write does not return the "compressed" or wrong value.

src/AspNetCore/Src/inprocessapplication.cxx
jkotalik: this is done, don't know why the diff is still there.

jkotalik:
@davidfowl I think I can solve the deadlock, but let me add that fix to a new PR instead of this one.

jkotalik:
🏓

src/AspNetCore/Src/forwardinghandler.cxx
```diff
@@ -1534,7 +1529,36 @@ REQUEST_NOTIFICATION_STATUS
...
+    // Application is null here? Can this happen for out of proc scenario?
+    if (m_pApplication->QueryConfig()->QueryHostingModel() == HOSTING_IN_PROCESS)
+    {
+        if (FAILED(hrCompletionStatus))
+        {
+            return RQ_NOTIFICATION_FINISH_REQUEST;
+        }
+        else
+        {
+            // For now wea re assuming we are in our own self contained box.
```
pan-wang: typo

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -189,22 +305,85 @@ http_flush_response_bytes(
...
+    _In_ IHttpContext* pHttpContext
+)
+{
+    if (!g_fWebSocketSupported)
+    {
+        return E_FAIL;
+    }
 
-IN_PROCESS_APPLICATION*  IN_PROCESS_APPLICATION::s_Application = NULL;
+    ((IHttpContext3*)pHttpContext)->EnableFullDuplex();
```
pan-wang: I think this will lead to crash on Win7 as the method cannot be found

src/AspNetCore/Src/forwardinghandler.cxx
```diff
@@ -1534,7 +1529,36 @@ REQUEST_NOTIFICATION_STATUS
...
+            if (application == NULL)
+            {
+                hr = E_FAIL;
+                return RQ_NOTIFICATION_FINISH_REQUEST;
+            }
+
+            if (application->QueryIsWebSocketsConnection())
+            {
+                // If this event occurs, we need to let the registered callback handler take care of this.
+                return RQ_NOTIFICATION_PENDING;
```
pan-wang: this will lead to leak of request as no one will return continue

src/AspNetCore/Src/inprocessapplication.cxx
jkotalik: Yeah these need to be put in a separate file eventually.

jkotalik: Next PR?

src/AspNetCore/Src/forwardinghandler.cxx
jkotalik: This is a safety check if the websocket connection causes an OnAsyncCompletion event to be triggered. Continue would eventually be called once the websocket connection is closed.

src/AspNetCore/Src/inprocessapplication.cxx
jkotalik: I'd rather keep the non websocket scenario the same across each platform. There should be no point in enabling duplex mode if we are only doing one read and write operation at a time.

src/AspNetCore/Src/inprocessapplication.cxx
jkotalik: Actually, this may fix all of our problems @davidfowl . Can't we just check in managed if there is a currently running async operation? If there is, we just say fAsync is false?

src/AspNetCore/Src/inprocessapplication.cxx
pan-wang: yes. I will do some test on win7 and solve it with next PR

pan-wang (approved):
let's commit this PR and solve the problems in next PR

PR #202, opened 2017-10-18
Adds Managed to Native abort logic.

jkotalik:


jkotalik:
Abort is a bit trickier here because after notifying IIS to abort, reads and writes will not fail by calling ReadEntityBody/WriteEntityChunks. Therefore, I add aborted to the stored context. @pan-wang and I will go through the native to managed cancellation later.

src/AspNetCore/Src/inprocessstoredcontext.cxx
```diff
@@ -32,6 +32,22 @@ IN_PROCESS_STORED_CONTEXT::QueryHttpContext(
     return m_pHttpContext;
 }
 
+VOID
+IN_PROCESS_STORED_CONTEXT::SetRequestAborted(
+    BOOL fRequestAborted
```
pan-wang: since you cannot change a request  back to un-aborted if it got aborted, no need for passing in the bool value

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -176,12 +201,26 @@ http_write_response_bytes(
     _In_ BOOL* pfCompletionExpected
 )
 {
+    HRESULT hr;
```
pan-wang: please initialize it to S_OK

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -176,12 +201,26 @@ http_write_response_bytes(
...
+        return hr;
+    }
+
+    if (pInProcessStoredContext->QueryRequstAborted())
+    {
+        *pfCompletionExpected = FALSE;
+        return E_ABORT;
+    }
+
     IHttpResponse *pHttpResponse = (IHttpResponse*)pHttpContext->GetResponse();
```
pan-wang: please declare the variables at the top of the function

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -46,13 +46,26 @@ http_get_raw_response(
     return pHttpContext->GetResponse()->GetRawHttpResponse();
 }
 
-EXTERN_C __MIDL_DECLSPEC_DLLEXPORT VOID http_set_response_status_code(
+EXTERN_C __MIDL_DECLSPEC_DLLEXPORT HRESULT http_set_response_status_code(
```
pan-wang: the signature may need to updated as you may also want to set sub-status 
https://support.microsoft.com/en-us/help/943891/the-http-status-code-in-iis-7-0--iis-7-5--and-iis-8-0

jkotalik: This is fixed, diff still showing up.

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -46,13 +46,27 @@ http_get_raw_response(
...
+    {
+        return hr;
+    }
+
+    if (pInProcessStoredContext->QueryRequstAborted())
+    {
+        return E_ABORT;
+    }
+
+    return pHttpContext->GetResponse()->SetStatus(statusCode, pszReason, subStatusCode);
```
Tratcher: What would happen if you called SetStatus after ResetConnection? How about WriteEntityChunks? Does it matter?
I wonder if we can skip all of this new API and call ResetConnection directly from managed.

src/AspNetCore/Src/inprocessstoredcontext.cxx
```diff
@@ -32,6 +32,21 @@ IN_PROCESS_STORED_CONTEXT::QueryHttpContext(
...
 
+VOID
+IN_PROCESS_STORED_CONTEXT::SetRequestAborted(
+)
+{
+    m_fRequestAborted = TRUE;
+}
+
+BOOL
+IN_PROCESS_STORED_CONTEXT::QueryRequstAborted(
```
Tratcher: Request

src/AspNetCore/Src/inprocessapplication.cxx
jkotalik: Calling SetStatus and WriteEntityChunks will succeed after calling ResetConnection. It is just being written to a local buffer. 
@pan-wang as part of the redesign of the pInvokes, will we be directly exposing HttpContext methods? If so, I'd agree it is better to have many of these calls in managed as currently most are wrappers for the native calls.

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -46,13 +46,27 @@ http_get_raw_response(
...
-    pHttpContext->GetResponse()->SetStatus(statusCode, pszReason);
+    HRESULT hr S_OK;
+    IN_PROCESS_STORED_CONTEXT* pInProcessStoredContext = NULL;
+    hr = IN_PROCESS_STORED_CONTEXT::GetInProcessStoredContext(pHttpContext, &pInProcessStoredContext);
+    if (FAILED(hr))
+    {
+        return hr;
+    }
+
+    if (pInProcessStoredContext->QueryRequestAborted())
```
pan-wang: current design looks a lit bit wired. The request is aborted by the managed layer. Why not immediately stop reading the request at the managed layer once the managed layer decides to abort the request instead of continue reading and checking it at the native layer.

jkotalik: Yeah I actually think I can remove these checks entirely. Writes and reads, if they occur, should succeed. The request will truly be aborted when the disconnect listener is fired.

src/AspNetCore/Src/managedexports.cxx
```diff
@@ -325,4 +339,15 @@ http_cancel_io(
     return pHttpContext->CancelIo();
 }
 
+EXTERN_C __MIDL_DECLSPEC_DLLEXPORT
+HRESULT
+http_abort_request(
```
davidfowl: http_reset_connection

src/AspNetCore/Src/managedexports.cxx
```diff
@@ -173,12 +177,22 @@ http_write_response_bytes(
     _In_ BOOL* pfCompletionExpected
 )
 {
-    IHttpResponse *pHttpResponse = (IHttpResponse*)pHttpContext->GetResponse();
+    HRESULT hr S_OK;
+    IHttpResponse *pHttpResponse = NULL;
+    IN_PROCESS_STORED_CONTEXT* pInProcessStoredContext = NULL;
```
pan-wang: no one reference to this stored context

jkotalik:
This will be merged pending #212

src/AspNetCore/Src/managedexports.cxx
```diff
@@ -158,19 +159,22 @@ http_read_request_bytes(
     _Out_ BOOL* pfCompletionPending
 )
 {
-    HRESULT hr;
+    HRESULT hr S_OK;
+    IN_PROCESS_STORED_CONTEXT* pInProcessStoredContext = NULL;
```
davidfowl: What is this for?

src/AspNetCore/Src/managedexports.cxx
```diff
@@ -158,19 +159,22 @@ http_read_request_bytes(
...
     if (pHttpContext == NULL)
     {
         return E_FAIL;
     }
+
     if (dwCbBuffer == 0)
     {
         return E_FAIL;
     }
-    IHttpRequest *pHttpRequest = (IHttpRequest*)pHttpContext->GetRequest();
```
davidfowl, line 171: What's up with these changes? They look unrelated?

jkotalik: This is just style in keeping the declarations at the beginning of the file.

jkotalik:
@pan-wang Let's coordinate this PR with the disconnect listener changes.

src/AspNetCore/Src/managedexports.cxx
```diff
@@ -197,12 +200,21 @@ http_write_response_bytes(
     _In_ BOOL* pfCompletionExpected
 )
 {
-    IHttpResponse *pHttpResponse = (IHttpResponse*)pHttpContext->GetResponse();
+    HRESULT hr S_OK;
```
pan-wang, line 203: where is "="?

src/AspNetCore/Src/managedexports.cxx
```diff
@@ -197,12 +200,21 @@ http_write_response_bytes(
...
+    IHttpResponse *pHttpResponse = NULL;
     BOOL fAsync = TRUE;
     BOOL fMoreData = TRUE;
     DWORD dwBytesSent = 0;
 
-    HRESULT hr = pHttpResponse->WriteEntityChunks(
+
+    if (pHttpContext == NULL)
+    {
+        return E_FAIL;
```
pan-wang, line 212: This null check is questionable as we control the stack unless you intentionally allow null input. Otherwise you can put null debug_assert to indicate that null input should not allowed.

this function can be simplified

src/AspNetCore/Src/managedexports.cxx
```diff
@@ -221,12 +233,12 @@ http_flush_response_bytes(
 )
 {
     IHttpResponse *pHttpResponse = (IHttpResponse*)pHttpContext->GetResponse();
-
+    HRESULT hr = S_OK;
```
pan-wang, line 236: since you already hardcoded the value, why do you still need those local variables

src/AspNetCore/Src/managedexports.cxx
```diff
@@ -349,4 +361,15 @@ http_cancel_io(
     return pHttpContext->CancelIo();
 }
 
+EXTERN_C __MIDL_DECLSPEC_DLLEXPORT
+HRESULT
+http_reset_connection(
+    _In_ IHttpContext* pHttpContext
```
pan-wang, line 367: not sure about why you want to return hardcoded hresult

jkotalik:
Closing for #227

PR #210, opened 2017-10-23
Expose VirtualDirectory to In Process mode.

jkotalik:
For https://github.com/aspnet/IISIntegration/issues/427

### DIFF
--- a/src/AspNetCore/Inc/aspnetcoreconfig.h
+++ b/src/AspNetCore/Inc/aspnetcoreconfig.h
@@ -140,6 +140,14 @@ public:
     }
 
     STRU*
+    QueryApplicationVirtualPath(
+        VOID
+    )
+    {
+        return &m_struApplicationVirtualPath;
+    }
+
+    STRU*
     QueryProcessPath(
         VOID
     )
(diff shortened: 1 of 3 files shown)


src/AspNetCore/Src/aspnetcoreconfig.cxx
```diff
@@ -470,4 +481,76 @@ ASPNETCORE_CONFIG::Populate(
     }
 
     return hr;
+}
+
+HRESULT
+ASPNETCORE_CONFIG::SetupAppPath(
+    IHttpContext*            pHttpContext,
```
pan-wang: it seems nobody references to pHttpContext as you passed in the siteconfigpath and resolved the path from it

Tratcher:
Is this the same value as is passed to out-of-proc via env var?

jkotalik:
I believe so. I guess we could just read the env variable?

jkotalik:
Though IMO it makes no sense to be reading env variables in process, as we have direct access to the structs.

src/AspNetCore/Inc/aspnetcoreconfig.h
```diff
@@ -222,6 +230,14 @@ class ASPNETCORE_CONFIG : IHttpStoredContext
         IHttpContext *pHttpContext
     );
 
+    HRESULT
+    SetupAppPath(
```
pan-wang: it seems no one uses this new method. you can delete it

pan-wang (approved):

PR #211, opened 2017-10-23
moving export methods to a standalone file

pan-wang:


src/AspNetCore/Src/managedexports.cxx
```diff
@@ -0,0 +1,328 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
```
jkotalik, line 1: Is it more appropriate to have these in an h file rather than cxx?

jkotalik (approved):
Thanks for doing that!

PR #212, opened 2017-10-24
Changes PostCompletion to handle OnAsyncCompletion after managed request has completed. 

jkotalik:
More stuff for the Async PR in IISIntegration: https://github.com/aspnet/IISIntegration/pull/447

### DIFF
--- a/src/AspNetCore/Src/inprocessapplication.cxx
+++ b/src/AspNetCore/Src/inprocessapplication.cxx
@@ -43,11 +43,7 @@ IN_PROCESS_APPLICATION::OnAsyncCompletion(
     {
         // means PostCompletion has been called and this is the associated callback.
         dwRequestNotificationStatus = pInProcessStoredContext->QueryAsyncCompletionStatus();
-        if (dwRequestNotificationStatus == RQ_NOTIFICATION_FINISH_REQUEST)
-        {
-            pHttpContext->DisableNotifications(RQ_END_REQUEST, 0);
-            // TODO cleanup whatever disconnect listener there is
-        }
+        // TODO cleanup whatever disconnect listener there is
         return dwRequestNotificationStatus;
     }
     else
--- a/src/AspNetCore/Src/inprocessapplication.cxx
+++ b/src/AspNetCore/Src/inprocessapplication.cxx
@@ -30,6 +30,7 @@ IN_PROCESS_APPLICATION::OnAsyncCompletion(
 {
     HRESULT hr;
     IN_PROCESS_STORED_CONTEXT* pInProcessStoredContext = NULL;
+    REQUEST_NOTIFICATION_STATUS dwRequestNotificationStatus = RQ_NOTIFICATION_CONTINUE;
 
     hr = IN_PROCESS_STORED_CONTEXT::GetInProcessStoredContext(pHttpContext, &pInProcessStoredContext);
     if (FAILED(hr))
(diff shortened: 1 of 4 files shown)


src/AspNetCore/Src/managedexports.cxx
```diff
@@ -56,9 +56,24 @@ EXTERN_C __MIDL_DECLSPEC_DLLEXPORT
 HRESULT
 http_post_completion(
```
davidfowl: I don't think this logic should be in post_completion

jkotalik:
@pan-wang We are going to be calling PostCompletion from managed to native. By calling this, OnAsyncCompletion will be triggered again by the proxymodule, which cannot call back into managed as managed will be disposing. To handle this, before calling post completion, we call http_set_completion_status() to set the status and that we have completed processing in managed. We then check in PostCompletion if the request has completed, and if so we return the status.

jkotalik:
This needs to be synced with https://github.com/aspnet/IISIntegration/pull/447

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -38,9 +39,22 @@ IN_PROCESS_APPLICATION::OnAsyncCompletion(
...
-    // Call the managed handler for async completion.
-    return m_AsyncCompletionHandler(pInProcessStoredContext->QueryManagedHttpContext(), hrCompletionStatus, cbCompletion);
+    else if (pInProcessStoredContext->QueryIsManagedRequestComplete())
+    {
+        // means PostCompletion has been called and this is the associated callback.
+        dwRequestNotificationStatus = pInProcessStoredContext->QueryAsyncCompletionStatus();
+        if (dwRequestNotificationStatus == RQ_NOTIFICATION_FINISH_REQUEST)
+        {
+            pHttpContext->DisableNotifications(RQ_END_REQUEST, 0);
+            // TODO cleanup whatever disconnect listener there is
```
davidfowl: Is there an issue for this?

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -38,9 +39,22 @@ IN_PROCESS_APPLICATION::OnAsyncCompletion(
...
-
-    // Call the managed handler for async completion.
-    return m_AsyncCompletionHandler(pInProcessStoredContext->QueryManagedHttpContext(), hrCompletionStatus, cbCompletion);
+    else if (pInProcessStoredContext->QueryIsManagedRequestComplete())
+    {
+        // means PostCompletion has been called and this is the associated callback.
+        dwRequestNotificationStatus = pInProcessStoredContext->QueryAsyncCompletionStatus();
+        if (dwRequestNotificationStatus == RQ_NOTIFICATION_FINISH_REQUEST)
+        {
+            pHttpContext->DisableNotifications(RQ_END_REQUEST, 0);
```
davidfowl: do we need this?

pan-wang: ANCM will not receive RQ_NOTIFICATION_FINISH_REQUEST as it did not register it.
I don't think we need it. Inprocess should always call m_AsyncCompletionHandler

src/AspNetCore/Src/inprocessstoredcontext.cxx
```diff
@@ -32,6 +34,38 @@ IN_PROCESS_STORED_CONTEXT::QueryHttpContext(
     return m_pHttpContext;
 }
 
+BOOL
+IN_PROCESS_STORED_CONTEXT::QueryIsManagedRequestComplete(
```
pan-wang, line 38: the native code ideally should not touch this stored managed context. Not sure why you need it

src/AspNetCore/Src/inprocessapplication.cxx
jkotalik: Alright I'll drop it.

src/AspNetCore/Src/inprocessapplication.cxx
jkotalik: I'll make a separate issue for adding the disconnect listener to managed and native.

src/AspNetCore/Src/inprocessstoredcontext.cxx
jkotalik: This doesn't touch the stored context. This is for knowing that an OnAsyncCompletion call was associated with PostCompletion(). If the context calls post completion, we should expect that the request is done and the managed request is being disposed.

jkotalik:
@pan-wang Is this one good to go after offline feedback?

pan-wang (approved):

PR #221, opened 2017-10-25
Standardizing managed exports to all return hresult

jkotalik:


src/AspNetCore/Src/managedexports.cxx
```diff
@@ -26,30 +26,37 @@ register_callbacks(
...
 )
 {
-    return pHttpContext->GetRequest()->GetRawHttpRequest();
+    *pHttpRequest = pHttpContext->GetRequest()->GetRawHttpRequest();
+    return S_OK;
 }
 
 EXTERN_C __MIDL_DECLSPEC_DLLEXPORT
-HTTP_RESPONSE*
+HRESULT
```
pan-wang, line 41: as this function always return s_ok, why not make it void

src/AspNetCore/Src/managedexports.cxx
```diff
@@ -110,17 +117,18 @@ http_set_managed_context(
 }
 
 EXTERN_C __MIDL_DECLSPEC_DLLEXPORT
-VOID
+HRESULT
```
pan-wang, line 120: what's the motivation for this change?

jkotalik: I wanted to standardize all pInvokes to return HRESULTs. In the future, if the implementation of the functions changes and suddenly we can return an invalid HRESULT, I don't want to have to make a second method.

jkotalik:
Closing for now as this may be unnecessary.

PR #223, opened 2017-10-27
client disconnect change

pan-wang:
initial check in. We can add managed callback later once we decides how to implement it

### DIFF
--- a/src/AspNetCore/Src/forwardinghandler.cxx
+++ b/src/AspNetCore/Src/forwardinghandler.cxx
@@ -1167,6 +1167,7 @@ FORWARDING_HANDLER::OnExecuteRequestHandler(
                 NULL,
                 L"InProcess Application");
         }
+        SetHttpSysDisconnectCallback();
         return  m_pApplication->ExecuteRequest(m_pW3Context);
     }
     case HOSTING_OUT_PROCESS:
@@ -1397,9 +1398,11 @@ Failure:
             &cchANCMHeader))) // first time failure
         {
             if (SUCCEEDED(hr = m_pW3Context->CloneContext(
-                CLONE_FLAG_BASICS | CLONE_FLAG_HEADERS | CLONE_FLAG_ENTITY,
-                &m_pChildRequestContext
-            )) &&
+                    CLONE_FLAG_BASICS  |
+                    CLONE_FLAG_HEADERS |
+                    CLONE_FLAG_ENTITY  |
+                    CLONE_FLAG_SERVER_VARIABLE,
+                    &m_pChildRequestContext)) &&
                 SUCCEEDED(hr = m_pChildRequestContext->SetServerVariable(
                     STR_ANCM_CHILDREQUEST,
                     L"1")) &&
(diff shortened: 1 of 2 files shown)


src/AspNetCore/Src/forwardinghandler.cxx
```diff
@@ -1397,9 +1398,11 @@ FORWARDING_HANDLER::OnExecuteRequestHandler(
             &cchANCMHeader))) // first time failure
         {
             if (SUCCEEDED(hr = m_pW3Context->CloneContext(
-                CLONE_FLAG_BASICS | CLONE_FLAG_HEADERS | CLONE_FLAG_ENTITY,
-                &m_pChildRequestContext
-            )) &&
+                    CLONE_FLAG_BASICS  |
+                    CLONE_FLAG_HEADERS |
+                    CLONE_FLAG_ENTITY  |
+                    CLONE_FLAG_SERVER_VARIABLE,
```
jkotalik, line 1404: Why do we need to add clone server variables?

src/AspNetCore/Src/forwardinghandler.cxx
```diff
@@ -1412,6 +1415,7 @@ FORWARDING_HANDLER::OnExecuteRequestHandler(
             {
                 if (!fCompletionExpected)
                 {
+                    m_pW3Context->SetRequestHandled();
```
jkotalik, line 1418: And here?

src/AspNetCore/Src/forwardinghandler.cxx
```diff
@@ -3141,45 +3145,54 @@ FORWARDING_HANDLER::TerminateRequest(
...
-
-    if (TlsGetValue(g_dwTlsIndex) != this)
+    if (m_pApplication->QueryConfig()->QueryHostingModel() == HOSTING_IN_PROCESS) 
     {
-        DBG_ASSERT(TlsGetValue(g_dwTlsIndex) == NULL);
-        AcquireSRWLockShared(&m_RequestLock);
-        TlsSetValue(g_dwTlsIndex, this);
-        DBG_ASSERT(TlsGetValue(g_dwTlsIndex) == this);
-        fAcquiredLock = TRUE;
+        //
```
jkotalik, line 3150: I'll need to verify that this change works with the abort changes, however this seems like it would work.

jkotalik (commented):
Doing some verification now. We should try to handle both flows for abort now (in this PR) rather than later.

jkotalik:
This looks good for now. Let's merge this for now so I can use it in managed.

jkotalik (approved):

src/AspNetCore/Src/forwardinghandler.cxx
pan-wang: we should. It was missing because we targeted on Vista in old depot which does not have this option.

src/AspNetCore/Src/forwardinghandler.cxx
pan-wang: this is standard IIS pattern since win7. We should follow it

PR #225, opened 2017-10-27
Fix the uninitialized g_hWinHttpModule global variable to avoid empty error messages

### LINKED ISSUE #224: The g_hWinHttpModule global variable is never assigned to
The `g_hWinHttpModule` global variable (declared in forwardinghandler.h) is never assigned to.

It is only ever used as an argument to `FormatMessage` function, which will fail to look up error message when `lpSource` parameter is not a valid module handle.

0xfeeddeadbeef:
Fixes #224

Feel free to discard this pull request, because I am not quite confident in my C++ skills — this might be completely wrong way to fix the issue. Also, I was not able to test it 😞 

- Assign module handle of winhttp.dll to `g_hWinHttpModule` in `RegisterModule` function.
 
- Lots of error codes between `WINHTTP_ERROR_BASE` and `WINHTTP_ERROR_LAST` do not
have corresponding entries in winhttp.dll message resource table; use defensive
coding to fall back to the default error message (`IDS_SERVER_ERROR`) when necessary.

For example, winhttp.dll on Windows Server 2012 R2 does not have message resource for `ERROR_WINHTTP_INVALID_HEADER`.

Running this code snippet will demonstrate that most valid error codes do not have corresponding message resources:

```c++
DWORD dwError = ERROR_SUCCESS;
HMODULE g_hWinHttpModule = GetModuleHandle(TEXT("winhttp.dll"));
TCHAR msgBuffer[0x10000 / sizeof(TCHAR)];  // Max size of buffer is 64K bytes

for (DWORD winHttpError = WINHTTP_ERROR_BASE + 1;
           winHttpError <= WINHTTP_ERROR_LAST;
           winHttpError++)
{
    memset(msgBuffer, 0, sizeof msgBuffer);

    if (FormatMessage(
        FORMAT_MESSAGE_IGNORE_INSERTS | FORMAT_MESSAGE_FROM_HMODULE,
        g_hWinHttpModule,
        winHttpError,
        0,
        msgBuffer,
        ARRAYSIZE(msgBuffer),
        NULL) == 0)
    {
        dwError = GetLastError();

        // "The system cannot find message text for message number 0x%1 in the message file for %2."
        if (dwError == ERROR_MR_MID_NOT_FOUND) {
            _ftprintf_s(stderr, _T("ERROR: %lu => Message not found!\r\n"), winHttpError);
        } else {
            _ftprintf_s(stderr, _T("ERROR: %lu => %lu\r\n"), winHttpError, dwError);
        }
    }
    else
    {
        _ftprintf_s(stdout, _T("MESSAGE: %lu => %s"), winHttpError, msgBuffer);
    }
}
```

### DIFF
--- a/src/AspNetCore/Src/dllmain.cpp
+++ b/src/AspNetCore/Src/dllmain.cpp
@@ -191,6 +191,7 @@ HRESULT
     g_pModuleId = pModuleInfo->GetId();
     g_pszModuleName = pModuleInfo->GetName();
     g_pHttpServer = pHttpServer;
+    g_hWinHttpModule = GetModuleHandle(TEXT("winhttp.dll"));
 
     //
     // WinHTTP does not create enough threads, ask it to create more.
--- a/src/AspNetCore/Src/forwardinghandler.cxx
+++ b/src/AspNetCore/Src/forwardinghandler.cxx
@@ -1447,9 +1447,8 @@ Failure:
     //
     pResponse->SetStatus(502, "Bad Gateway", 3, hr);
 
-    if (hr > HRESULT_FROM_WIN32(WINHTTP_ERROR_BASE) &&
-        hr <= HRESULT_FROM_WIN32(WINHTTP_ERROR_LAST))
-    {
+    if (!(hr > HRESULT_FROM_WIN32(WINHTTP_ERROR_BASE) &&
+          hr <= HRESULT_FROM_WIN32(WINHTTP_ERROR_LAST)) ||
 #pragma prefast (suppress : __WARNING_FUNCTION_NEEDS_REVIEW, "Function and parameters reviewed.")
         FormatMessage(
             FORMAT_MESSAGE_IGNORE_INSERTS | FORMAT_MESSAGE_FROM_HMODULE,
@@ -1458,9 +1457,7 @@ Failure:
             0,
             strDescription.QueryStr(),
             strDescription.QuerySizeCCH(),
-            NULL);
-    }
-    else
+            NULL) == 0)
     {
         LoadString(g_hModule,
             IDS_SERVER_ERROR,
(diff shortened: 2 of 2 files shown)


ghost:
Hi, I am closing and re-opening this PR to bump the CLA service. Sorry for the inconvenience!

src/AspNetCore/Src/forwardinghandler.cxx
```diff
@@ -1451,14 +1451,20 @@ FORWARDING_HANDLER::OnExecuteRequestHandler(
...
             FORMAT_MESSAGE_IGNORE_INSERTS | FORMAT_MESSAGE_FROM_HMODULE,
             g_hWinHttpModule,
             HRESULT_CODE(hr),
             0,
             strDescription.QueryStr(),
             strDescription.QuerySizeCCH(),
-            NULL);
+            NULL) == 0)
+            {
+                LoadString(g_hModule,
```
jkotalik: The same LoadString function is called below. I think you can simplify the logic to be have FormatMessage in the if statement.

src/AspNetCore/Src/dllmain.cpp
```diff
@@ -191,6 +191,7 @@ HRESULT
     g_pModuleId = pModuleInfo->GetId();
     g_pszModuleName = pModuleInfo->GetName();
     g_pHttpServer = pHttpServer;
+    g_hWinHttpModule = GetModuleHandle(TEXT("winhttp.dll"));
```
jkotalik, line 194: @pan-wang is this correct?

jkotalik (commented):
Thanks for the contribution. Good find! Awaiting @pan-wang approval/feedback.

pan-wang:
Nice catch. It is a tricky part. In the past we copied WinHttp.h and added a lot defines such as  ERROR_WINHTTP_INVALID_HEADER to achieve compatibility. As we cannot use this customized header file, we need revisit it and make sure nothing is missed.

0xfeeddeadbeef:
@jkotalik I've force-pushed changes based on your code review. As a result, your original comment is gone. Sorry.

src/AspNetCore/Src/forwardinghandler.cxx
0xfeeddeadbeef: @jkotalik Done.

src/AspNetCore/Src/forwardinghandler.cxx
```diff
@@ -1447,9 +1447,8 @@ FORWARDING_HANDLER::OnExecuteRequestHandler(
     //
     pResponse->SetStatus(502, "Bad Gateway", 3, hr);
 
-    if (hr > HRESULT_FROM_WIN32(WINHTTP_ERROR_BASE) &&
-        hr <= HRESULT_FROM_WIN32(WINHTTP_ERROR_LAST))
-    {
+    if (!(hr > HRESULT_FROM_WIN32(WINHTTP_ERROR_BASE) &&
+          hr <= HRESULT_FROM_WIN32(WINHTTP_ERROR_LAST)) ||
```
pan-wang, line 1451: The code logic was change here. In the old code if FormatMessage failed, we will not try LoadString. 
This change make sense to me.

pan-wang (approved):

PR #227, opened 2017-10-30
Adds disconnect callback. 

jkotalik:
Can't merge until we figure out why disconnect is being triggered by shutdown.


src/AspNetCore/Src/managedexports.cxx
```diff
@@ -355,4 +357,13 @@ http_cancel_io(
     return pHttpContext->CancelIo();
 }
 
+EXTERN_C __MIDL_DECLSPEC_DLLEXPORT
+VOID
+http_reset_connection(
```
muratg, line 370: Did you add this so it's easier to reset connection?  Or will there be additions to this function later on?

jkotalik: This is it. Ideally these functions should be very light weight wrappers around the native objects. We could directly expose each of the functions, but as most methods do more than just one line, I've kept them this way.

jkotalik: Also this function is used for request aborted.

src/AspNetCore/Src/forwardinghandler.cxx
```diff
@@ -3147,9 +3147,10 @@ FORWARDING_HANDLER::TerminateRequest(
 {
     if (m_pApplication->QueryConfig()->QueryHostingModel() == HOSTING_IN_PROCESS) 
     {
-        //
-        // Todo: need inform managed layer to abort the request
-        //
+        // TODO Terminate Request is being called on Shutdown of IISExpress, causing 
+        // disconnect to be called at the wrong time.
+        IN_PROCESS_APPLICATION* application = (IN_PROCESS_APPLICATION*)m_pApplication;
```
davidfowl, line 3152: Can this fire after the request is over?

jkotalik: I don't believe so now. With this change: https://github.com/aspnet/AspNetCoreModule/pull/227/files/5dd3a7777d1d021683c447532dcc0868032a9dc5#diff-6752c6fb79cdaf9a4ff73a601b7b4e5fR93, we now will unregister the disconnect context from the request.

src/AspNetCore/Inc/inprocessapplication.h
```diff
@@ -53,8 +54,14 @@ class IN_PROCESS_APPLICATION : public APPLICATION
...
     LoadManagedApplication(
-            VOID
-        );
+        VOID
+    );
+    
+    HRESULT
+    OnClientDisconnect(
+        IHttpContext* pHttpContext,
+        BOOL fClientDisconnected
```
pan-wang: why do we need thisfClientDisconnected?  We can assume client is disconnected when this function is called

pan-wang, line 63: I don't think we need this fClientDisconnected, since this value is not flowed to managed layer.

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -651,6 +652,31 @@ IN_PROCESS_APPLICATION::ExecuteApplication(
     return hr;
 }
 
+HRESULT
```
pan-wang, line 655: No one will use this returned HRESULT, as we are in a callback and cannot interrupt the main thread.

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -651,6 +652,31 @@ IN_PROCESS_APPLICATION::ExecuteApplication(
...
+    if ( m_ClientDisconnectHandler == NULL || !fClientDisconnected)
+    {
+        return E_FAIL;
+    }
+
+    hr = IN_PROCESS_STORED_CONTEXT::GetInProcessStoredContext( pHttpContext, &pInProcessStoredContext );
+    if ( FAILED( hr ) )
+    {
+        // Finish the request as we couldn't get the callback
+        pHttpContext->GetResponse()->SetStatus( 500, "Internal Server Error", 19, hr );
```
pan-wang, line 673: Most likely this function will be called ion another thread. Setting response status in failure case will not take effect as the main thread still process the request and will set http response. I think this function can be simplified.
We also need some kind interlockexchange mechanism to prevent main thread from the race condition. I would suggest to process against forwarding handler. That is main thread and the callback thread should set/check a flag in forwarding handler to hold the stored context.

davidfowl: @pan-wang what thread does this fire on? Some random http.sys thread? If yes, how do we know the pHttpContext is still valid?

pan-wang: Some callback thread from HTTP.SYS. this is why we need a lock/flag mechanism to ensure the httpcontext is still valid.  For example,  before finish the request,  we need unregistered the callback and set a flag in forwarding handler.  The disconnect callback need to check the flag and set it (if it was not set yet) to prevent main thread release the request/forwarding handler object. This is a very dangerous code path which can easily cause AV and thus security issue.

davidfowl: Agreed. Do we have similar issues with the out of process code path as well?

jkotalik:
Will reopen once shim is in.

PR #228, opened 2017-10-31
Remove header serialization in favor of setting IIS struct directly.

jkotalik:
Response to https://github.com/aspnet/IISIntegration/pull/465.

### DIFF
--- a/src/AspNetCore/Src/managedexports.cxx
+++ b/src/AspNetCore/Src/managedexports.cxx
@@ -357,7 +357,7 @@ http_cancel_io(
 
 EXTERN_C __MIDL_DECLSPEC_DLLEXPORT
 HRESULT
-http_set_header(
+http_response_set_header(
     _In_ IHttpContext* pHttpContext,
     _In_ PCSTR pszHeaderName,
     _In_ PCSTR pszHeaderValue,
--- a/src/AspNetCore/Src/managedexports.cxx
+++ b/src/AspNetCore/Src/managedexports.cxx
@@ -355,4 +355,17 @@ http_cancel_io(
     return pHttpContext->CancelIo();
 }
 
+EXTERN_C __MIDL_DECLSPEC_DLLEXPORT
+HRESULT
+http_set_header(
+    _In_ IHttpContext* pHttpContext,
+    _In_ PCSTR pszHeaderName,
+    _In_ PCSTR pszHeaderValue,
+    _In_ USHORT usHeaderValueLength,
+    _In_ BOOL  fReplace
+)
+{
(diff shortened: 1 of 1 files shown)


src/AspNetCore/Src/managedexports.cxx
```diff
@@ -355,4 +355,17 @@ http_cancel_io(
...
+HRESULT
+http_response_set_header(
+    _In_ IHttpContext* pHttpContext,
+    _In_ PCSTR pszHeaderName,
+    _In_ PCSTR pszHeaderValue,
+    _In_ USHORT usHeaderValueLength,
+    _In_ BOOL  fReplace
+)
+{
+    return pHttpContext->GetResponse()->SetHeader( pszHeaderName, pszHeaderValue, usHeaderValueLength, fReplace );
```
Tratcher, line 368: There's an indexed overload you can use for known headers so you don't have to serialize and pin their names.
https://msdn.microsoft.com/en-us/library/ms694295(v=vs.90).aspx

Tratcher: Is http_response_set_header even needed? can't the managed code call SetHeader directly?

Tratcher: The same for http_set_response_status_code

jkotalik: The managed code can call SetHeader, but it still needs to do a pinvoke. Either way we need to access the native iis http context. At this point I am keen on not changing it.

jkotalik: I think we will need to add two native methods to handle this.

pan-wang: will you set known header and unknown header differently at managed layer? IIS SetHeader api does not distinguish known header and unknown header here. 
internally, IIS SetHeader will calculate the ID of the string input header name to determine it is known one or unknown one. Even given header index, IIS still check whether the header needs to be treated as unknown. I am not sure about the performance difference here. If managed code needs to do some calculation to determine whether the header is known or unknown. You may want to simplify the code to only call IIS SetHeader with string header value.

Tratcher: The known header API avoids having the managed layer serialize extra strings. As these are the most common headers, that's a lot of allocation, encoding, and pinning we can avoid.

Tratcher (changes requested):
We still don't have a good reason for adding these new methods when the managed layer can p-invoke SetHeader directly. Is there a reason?

jkotalik:
After offline discussion, we are on board with staying with this style. @pan-wang this good to go.

pan-wang (approved):

PR #229, opened 2017-11-02
Jhkim/fix testcodeissues

jhkimnew:
This is for fixing some minor test code issues.


test/AspNetCoreModule.TestSites.Standard/Startup.cs
```diff
@@ -318,6 +318,31 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
                             response += de.Key + ":" + de.Value + "<br/>";
                         }
                     }
+
```
pan-wang, line 321: you can use query string in url for better test maintenance. You can do it later

pan-wang (approved):

PR #230, opened 2017-11-02
Add location information to ANCM package

pakrym:
So package consumers have an easy way of finding dlls


Tratcher (approved):

muratg (commented):
Post check-in :shipit:  👍

PR #232, opened 2017-11-03
redirect stdout to log file + minor change to allow debug build

pan-wang:
For inprocess scenario, we redirect the stdout and stderr to a file handle using SetStdHandle API.
We always redirect the stderr even though user does not enable logging in web.config. This will help us to capture the start failure of managed application. Empty logging file will be deleted if logging is not enabled.
some small changes for out-of-process logging to make it consistent with inprocess logging.
A small change in project file to allow debug build.

### DIFF
--- a/src/AspNetCore/AspNetCore.vcxproj
+++ b/src/AspNetCore/AspNetCore.vcxproj
@@ -103,6 +103,7 @@
       <PrecompiledHeaderOutputFile>$(IntDir)$(TargetName).pch</PrecompiledHeaderOutputFile>
       <AdditionalIncludeDirectories>..\IISLib;.\Inc</AdditionalIncludeDirectories>
       <DebugInformationFormat>ProgramDatabase</DebugInformationFormat>
+      <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
     </ClCompile>
     <Link>
       <SubSystem>Windows</SubSystem>
--- a/src/AspNetCore/Src/aspnetcoreconfig.cxx
+++ b/src/AspNetCore/Src/aspnetcoreconfig.cxx
@@ -48,6 +48,114 @@ ASPNETCORE_CONFIG::~ASPNETCORE_CONFIG()
 }
 
 HRESULT
+ASPNETCORE_CONFIG::CreateLogFile(
+    _In_  BOOL    fIgnoreDisableInConfig,
+    _Out_ STRU*   pstruFullFileName,
+    _Out_ HANDLE* pHandle
+)
+{
+    HRESULT hr = HRESULT_FROM_NT(ERROR_NOT_SUPPORTED); // Assume log is not enabled
+    SECURITY_ATTRIBUTES     saAttr = { 0 };
+    SYSTEMTIME              systemTime;
+    STRU    struPath;
+
+    DBG_ASSERT(pstruFullFileName);
+    DBG_ASSERT(pHandle);
+
+    *pHandle = INVALID_HANDLE_VALUE;
+    if (fIgnoreDisableInConfig || m_fStdoutLogEnabled)
+    {
+        STRU   struTmp;
+        STRU   struLogPath;
+        WCHAR* strSegment = NULL;
+        WCHAR* strRestSegments = NULL;
+        hr = struTmp.Copy(m_struStdoutLogFile.QueryStr());
+        if (FAILED(hr))
+        {
+            goto Finished;
+        }
+
+        //
+        // Check whether the (first) segment exists based on configuration element 'stdoutLogFile'
+        // If not, create that folder. We only want make the default '.\logs\stdout' work.
+        // No recursive folder creation
+        //
+        strSegment = wcstok_s(struTmp.QueryStr(), L"\\", &strRestSegments);
+        if (strSegment != NULL && (strSegment[0] == L'.') && strSegment[1] == L'\0')
+        {
+            struLogPath.Append(L".\\");
+            strSegment = wcstok_s(NULL, L"\\", &strRestSegments);
+            if (strSegment != NULL)
+            {
+                struLogPath.Append(strSegment);
+            }
+            strSegment = struLogPath.QueryStr();
+        }
+        if (strSegment != NULL)
+        {
+            hr = PATH::ConvertPathToFullPath(
+                strSegment,
+                m_struApplicationFullPath.QueryStr(),
+                &struPath);
+            if (FAILED(hr))
+            {
+                goto Finished;
+            }
+            if (!CreateDirectory(struPath.QueryStr(), NULL) &&
+                ERROR_ALREADY_EXISTS != GetLastError())
+            {
+                hr = HRESULT_FROM_WIN32(GetLastError());
+                goto Finished;
+            }
+        }
+        struPath.Reset();
+        hr = PATH::ConvertPathToFullPath(
+            m_struStdoutLogFile.QueryStr(),
+            m_struApplicationFullPath.QueryStr(),
+            &struPath);
+        if (FAILED(hr))
+        {
+            goto Finished;
+        }
+
+        GetSystemTime(&systemTime);
+        hr = pstruFullFileName->SafeSnwprintf(L"%s_%d%02d%02d%02d%02d%02d_%d.log",
+                struPath.QueryStr(),
+                systemTime.wYear,
+                systemTime.wMonth,
+                systemTime.wDay,
+                systemTime.wHour,
+                systemTime.wMinute,
+                systemTime.wSecond,
+                GetCurrentProcessId());
+        if (FAILED(hr))
+        {
+           goto Finished;
+        }
+
+        saAttr.nLength = sizeof(SECURITY_ATTRIBUTES);
+        saAttr.bInheritHandle = TRUE;
+        saAttr.lpSecurityDescriptor = NULL;
+
+        *pHandle = CreateFileW(pstruFullFileName->QueryStr(),
+                FILE_WRITE_DATA,
+                FILE_SHARE_READ,
+                &saAttr,
+                CREATE_ALWAYS,
+                FILE_ATTRIBUTE_NORMAL,
+                NULL);
+        if (*pHandle == INVALID_HANDLE_VALUE)
+        {
+            hr = HRESULT_FROM_WIN32(GetLastError());
+            goto Finished;
+        }
+        hr = S_OK;
+    }
+Finished:
+    return hr;
+}
+
+HRESULT
 ASPNETCORE_CONFIG::GetConfig(
     _In_  IHttpContext            *pHttpContext,
     _Out_ ASPNETCORE_CONFIG     **ppAspNetCoreConfig
--- a/src/AspNetCore/Src/forwardinghandler.cxx
+++ b/src/AspNetCore/Src/forwardinghandler.cxx
@@ -4,7 +4,8 @@
 #include "precomp.hxx"
 #include <dbgutil.h>
 #include <comdef.h>
-
+#include <io.h>
+#include <stdio.h>
 
 // Just to be aware of the FORWARDING_HANDLER object size.
 C_ASSERT(sizeof(FORWARDING_HANDLER) <= 632);
@@ -1142,7 +1143,8 @@ FORWARDING_HANDLER::OnExecuteRequestHandler(
     {
     case HOSTING_IN_PROCESS:
     {
-
+        // Set Stdout
+        ((IN_PROCESS_APPLICATION*)m_pApplication)->SetStdOut();
         hr = ((IN_PROCESS_APPLICATION*)m_pApplication)->LoadManagedApplication();
         if (FAILED(hr))
         {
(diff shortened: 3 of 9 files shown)


pan-wang:
need to test on App Service to make sure no blocked API was used

davidfowl:
@pan-wang can you write a summary of the change in the PR description?

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -132,6 +135,125 @@ IN_PROCESS_APPLICATION::FindDotNetFolders(
...
+                    hr = m_pConfiguration->CreateLogFile(TRUE, &m_struLogFilePath, &m_hLogFileHandle);
+                    if (FAILED(hr))
+                    {
+                        goto Finished;
+                    }
+
+                    // The process has console, e.g., IIS Express scenario
+                    CloseHandle(m_hLogFileHandle);
+                    m_hLogFileHandle = INVALID_HANDLE_VALUE;
+                    if (_wfopen_s(&m_pStdFile, m_struLogFilePath.QueryStr(), L"w") == 0)
```
davidfowl, line 199: Does this force all logs (including the IIS Express logs) to go to the console?

pan-wang: yes. IISExpress directly write to console. Managed app also write to console. We cannot distinguish the sources

davidfowl: What happened to the idea of passing a file handle to the managed process?

pan-wang:
I strongly suspect whether it will work and the scope of change.  In the out of process scenario,  ANCM resets the console handle to a file when creating the child process and periodically flushs the buffet to the file.  It seems ANCM controls the console output and the managed layer never uses the file handle.





On 2017-11-04 23:28 , David Fowler<mailto:notifications@github.com> Wrote:


@davidfowl commented on this pull request.

________________________________

In src/AspNetCore/Src/inprocessapplication.cxx<https://github.com/aspnet/AspNetCoreModule/pull/232#discussion_r148949078>:

> +                }
+            }
+            else
+            {
+                if (m_pConfiguration->QueryStdoutLogEnabled())
+                {
+                    hr = m_pConfiguration->CreateLogFile(TRUE, &m_struLogFilePath, &m_hLogFileHandle);
+                    if (FAILED(hr))
+                    {
+                        goto Finished;
+                    }
+
+                    // The process has console, e.g., IIS Express scenario
+                    CloseHandle(m_hLogFileHandle);
+                    m_hLogFileHandle = INVALID_HANDLE_VALUE;
+                    if (_wfopen_s(&m_pStdFile, m_struLogFilePath.QueryStr(), L"w") == 0)


What happened to the idea of passing a file handle to the managed process?

src/AspNetCore/AspNetCore.vcxproj
```diff
@@ -103,6 +103,7 @@
       <PrecompiledHeaderOutputFile>$(IntDir)$(TargetName).pch</PrecompiledHeaderOutputFile>
       <AdditionalIncludeDirectories>..\IISLib;.\Inc</AdditionalIncludeDirectories>
       <DebugInformationFormat>ProgramDatabase</DebugInformationFormat>
+      <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
```
jkotalik, line 106: Intended change?

src/AspNetCore/Src/aspnetcoreconfig.cxx
```diff
@@ -48,6 +48,114 @@ ASPNETCORE_CONFIG::~ASPNETCORE_CONFIG()
...
+        hr = struTmp.Copy(m_struStdoutLogFile.QueryStr());
+        if (FAILED(hr))
+        {
+            goto Finished;
+        }
+
+        //
+        // Check whether the (first) segment exists based on configuration element 'stdoutLogFile'
+        // If not, create that folder. We only want make the default '.\logs\stdout' work.
+        // No recursive folder creation
```
jkotalik, line 81: Why don't we want to support recursive folder creation?

src/AspNetCore/Src/aspnetcoreconfig.cxx
```diff
@@ -48,6 +48,114 @@ ASPNETCORE_CONFIG::~ASPNETCORE_CONFIG()
...
+        strSegment = wcstok_s(struTmp.QueryStr(), L"\\", &strRestSegments);
+        if (strSegment != NULL && (strSegment[0] == L'.') && strSegment[1] == L'\0')
+        {
+            struLogPath.Append(L".\\");
+            strSegment = wcstok_s(NULL, L"\\", &strRestSegments);
+            if (strSegment != NULL)
+            {
+                struLogPath.Append(strSegment);
+            }
+            strSegment = struLogPath.QueryStr();
```
jkotalik, line 92: For .\logs\stdout, wouldn't this just output .\logs ? Will the output file be called stdout without an extension?

src/AspNetCore/Src/forwardinghandler.cxx
```diff
@@ -4,7 +4,8 @@
 #include "precomp.hxx"
 #include <dbgutil.h>
 #include <comdef.h>
-
+#include <io.h>
```
jkotalik: Do we need these includes?

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -4,6 +4,7 @@
 #include "precomp.hxx"
 #include <algorithm>
 
+
```
jkotalik: nit remove newline.

src/AspNetCore/Src/aspnetcoreconfig.cxx
muratg: cc @Tratcher

src/AspNetCore/AspNetCore.vcxproj
pan-wang: yes. Otherwise the resulting dll will have dependency on VS.

src/AspNetCore/Src/aspnetcoreconfig.cxx
pan-wang: We could. I did not see strong customer's ask and don't want to put more code to handle it. (more testing cases). We can do it in the future there is strong ask for this feature

src/AspNetCore/Src/aspnetcoreconfig.cxx
Tratcher: See https://github.com/aspnet/AspNetCoreModule/pull/232/files#diff-acdf64ca39d116f420e79aab69bed017R122

pan-wang: stdout is just a prefix. The whole file name will be stdout_<time><current-process-id>.log. this is consistent with old code

src/AspNetCore/Src/inprocessapplication.cxx
Tratcher: Who would accept it and re-route to it? UseIISIntegration? That's too late to capture a lot of common failures.

src/AspNetCore/Src/forwardinghandler.cxx
pan-wang: io.h is for _dup2(). 
stdio.h is or SetStdHandle()

src/AspNetCore/Src/inprocessapplication.cxx
pan-wang: done

src/AspNetCore/Src/forwardinghandler.cxx
jkotalik: We already include precomp though.

src/AspNetCore/AspNetCore.vcxproj
jkotalik: Only for debug builds, correct?

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -132,6 +134,125 @@ IN_PROCESS_APPLICATION::FindDotNetFolders(
...
+                    goto Finished;
+                }
+                //
+                // SetStdHandle works as w3wp does not have Console
+                // Current process does not have a console
+                //
+                SetStdHandle(STD_ERROR_HANDLE, m_hLogFileHandle);
+                if (m_pConfiguration->QueryStdoutLogEnabled())
+                {
+                    SetStdHandle(STD_OUTPUT_HANDLE, m_hLogFileHandle);
```
jkotalik, line 171: These comments lead me to think that SetStdHandle doesn't work, right?

pan-wang: No. SetStdHandle works for w3wp scenario.

src/AspNetCore/AspNetCore.vcxproj
pan-wang: yes. this will still generate debug build but does not support VS on time code change

jkotalik:
@pan-wang have you gotten the chance to check if these calls work with azure?

src/AspNetCore/Src/aspnetcoreconfig.cxx
```diff
@@ -48,6 +48,114 @@ ASPNETCORE_CONFIG::~ASPNETCORE_CONFIG()
 }
 
 HRESULT
+ASPNETCORE_CONFIG::CreateLogFile(
```
jkotalik, line 51: I don't see why this function should belong to aspnetcoreconfig. Consider moving it to a shared location and passing in the m_struStdoutLogFile and m_fStdoutLogEnabled.

src/AspNetCore/Src/serverprocess.cxx
```diff
@@ -1189,121 +1189,84 @@ SERVER_PROCESS::SetupStdHandles(
...
 {
-    SECURITY_ATTRIBUTES     saAttr = {0};
     HRESULT                 hr = S_OK;
     SYSTEMTIME              systemTime;
-    STRU                    struLogFileName;
     BOOL                    fStdoutLoggingFailed = FALSE;
     STRU                    strEventMsg;
     LPCWSTR                 apsz[1];
-    STRU                    struAbsLogFilePath;
+    ASPNETCORE_CONFIG*      pConfig = NULL;
```
jkotalik, line 1197: I don't like that we query the config here. Most of the fields in config are passed in Initialize().

src/AspNetCore/Src/inprocessapplication.cxx
jkotalik: Should we remove the commented out code though?

src/AspNetCore/Src/aspnetcoreconfig.cxx
pan-wang: yes. I did thought about it. Ideally, it should be a utility function

src/AspNetCore/Src/inprocessapplication.cxx
pan-wang: no. This is just comment is for maintenance purpose saying this method won't work. No need to try them later.

src/AspNetCore/Src/serverprocess.cxx
pan-wang: We could pass in the configuration object but it requires change amount of code. Since we already have the request object, it's arguable that passing is better than querying.

src/AspNetCore/Src/inprocessapplication.cxx
jkotalik: Let's move the comment before the call to SetStdHandle. Also change the comment to say "These other options do not work on azure or don't capture logging early enough"

src/AspNetCore/Src/inprocessapplication.cxx
```diff
@@ -132,6 +134,125 @@ IN_PROCESS_APPLICATION::FindDotNetFolders(
...
+                    m_hLogFileHandle = INVALID_HANDLE_VALUE;
+                    if (_wfopen_s(&m_pStdFile, m_struLogFilePath.QueryStr(), L"w") == 0)
+                    {
+                        // known issue: error info may not be capture when process crashes during buffering
+                        // even we disabled FILE buffering
+                        setvbuf(m_pStdFile, NULL, _IONBF, 0);
+                        _dup2(_fileno(m_pStdFile), _fileno(stdout));
+                        _dup2(_fileno(m_pStdFile), _fileno(stderr));
+                    }
+                    // not work for console scenario
```
jkotalik, line 207: Same here.

src/AspNetCore/Src/serverprocess.cxx
```diff
@@ -1992,6 +1955,21 @@ SERVER_PROCESS::~SERVER_PROCESS()
         m_hStdoutHandle = NULL;
     }
 
+    if (!m_fStdoutLogEnabled && !m_struFullLogFile.IsEmpty())
```
jkotalik, line 1958: Shouldn't this be 
```c++
if (m_fStdoutLogEnabled && !m_struFullLogFile.IsEmpty())

```

jkotalik (approved):
I'll add an issue to remove CreateLogFile from the aspnetcoreconfig. Also, a bit of cleanup on the comments

pan-wang:
no need for this PR anymore after refactoring

PR #233, opened 2017-11-06
Test: Added a new MiddleWare to test the Gracefulshutdown message

jhkimnew:
Added a new MiddleWare to process the graceful shutdown message before IISMiddleware in order to test the scenario of aspnetcore apps which does not handle graceful shutdown.

### DIFF
--- a/test/AspNetCoreModule.TestSites.Standard/Program.cs
+++ b/test/AspNetCoreModule.TestSites.Standard/Program.cs
@@ -2,13 +2,13 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using Microsoft.AspNetCore.Hosting;
-using Microsoft.AspNetCore.Server.Kestrel.Https;
 using Microsoft.Extensions.Configuration;
 using System;
 using System.IO;
 using System.Security.Cryptography.X509Certificates;
 using System.Threading;
-using System.Net;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace AspnetCoreModule.TestSites.Standard
 {
@@ -88,8 +88,18 @@ namespace AspnetCoreModule.TestSites.Standard
             else
             {
                 builder = new WebHostBuilder()
+                    .ConfigureServices(services =>
+                    {
+                        const string PairingToken = "TOKEN";
+                        string paringToken = builder.GetSetting(PairingToken) ?? Environment.GetEnvironmentVariable($"ASPNETCORE_{PairingToken}");
+                        services.AddSingleton<IStartupFilter>(
+                            new IISSetupFilter(
+                                paringToken,
+                                new Microsoft.AspNetCore.Http.PathString(null) // this parameter is not used yet
+                            )
+                        );
+                    })
                     .UseConfiguration(config)
-                    .UseIISIntegration()
                     .UseStartup<Startup>();
             }
 
@@ -139,7 +149,7 @@ namespace AspnetCoreModule.TestSites.Standard
             );
             try
             {
-                host.Run();                
+                host.Run();
             }
             catch
             {
--- a/test/AspNetCoreModule.TestSites.Standard/Program.cs
+++ b/test/AspNetCoreModule.TestSites.Standard/Program.cs
@@ -93,10 +93,7 @@ namespace AspnetCoreModule.TestSites.Standard
                         const string PairingToken = "TOKEN";
                         string paringToken = builder.GetSetting(PairingToken) ?? Environment.GetEnvironmentVariable($"ASPNETCORE_{PairingToken}");
                         services.AddSingleton<IStartupFilter>(
-                            new IISSetupFilter(
-                                paringToken,
-                                new Microsoft.AspNetCore.Http.PathString(null) // this parameter is not used yet
-                            )
+                            new IISSetupFilter(paringToken)
                         );
                     })
                     .UseConfiguration(config)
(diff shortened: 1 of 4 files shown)


test/AspNetCoreModule.TestSites.Standard/Program.cs
```diff
@@ -88,8 +88,18 @@ public static void Main(string[] args)
...
             {
                 builder = new WebHostBuilder()
+                    .ConfigureServices(services =>
+                    {
+                        const string PairingToken = "TOKEN";
+                        string paringToken = builder.GetSetting(PairingToken) ?? Environment.GetEnvironmentVariable($"ASPNETCORE_{PairingToken}");
+                        services.AddSingleton<IStartupFilter>(
+                            new IISSetupFilter(
+                                paringToken,
+                                new Microsoft.AspNetCore.Http.PathString(null) // this parameter is not used yet
```
Tratcher: Then remove it?

test/AspNetCoreModule.TestSites.Standard/TestMiddleWareBeforeIISMiddleWare.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            this.next = next;
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            // if the given request is shutdown message from ANCM and the value of GracefulShutdown environment variable is set,
+            // the shutdown message is handled by this middleware instead of IISMiddleware.
+
+            if (HttpMethods.IsPost(context.Request.Method) &&
```
Tratcher, line 27: There's also a request path

jhkimnew: Updated after CR from @Tratcher

test/AspNetCoreModule.TestSites.Standard/TestMiddleWareBeforeIISMiddleWare.cs
```diff
@@ -0,0 +1,54 @@
...
+                context.Request.Path.ToString().EndsWith("/iisintegration") &&
```
Tratcher, line 28: EndsWith? Why not just Path.Equals?

Tratcher (approved):

test/AspNetCoreModule.TestSites.Standard/TestMiddleWareBeforeIISMiddleWare.cs
jhkimnew: That is because this test app is not placed on the root application. It is placed on a child application on IIS websites. In that case, the Request Path will look like "/<childappName/iisintegration.

test/AspNetCoreModule.TestSites.Standard/Startup.cs
```diff
@@ -344,20 +344,6 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
                         }
                     }
                 }
-
-                // Handle shutdown event from ANCM
-                if (HttpMethods.IsPost(context.Request.Method) &&
```
jkotalik: Why do we not need this check anymore?

test/AspNetCoreModule.TestSites.Standard/TestMiddleWareBeforeIISMiddleWare.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                string shutdownMode = Environment.GetEnvironmentVariable("GracefulShutdown");
+                if (!string.IsNullOrWhiteSpace(shutdownMode) && shutdownMode.ToLower().StartsWith("disabled"))
+                {
+                    //ignore shutdown Message returning 200 instead of 202 because the gracefulshutdown is disabled
+                    context.Response.StatusCode = StatusCodes.Status200OK;
+                    await context.Response.WriteAsync("Called ShutdownMessage with disabled of GracefulShutdown");
+                    return;
+                }
+                // The below code might be required for old version of AspnetCore in case IISMiddleware does not handle the graceful shutdown message.
```
jkotalik: This shouldn't be commented out like this. We don't support this scenario, correct (1.0 ANCM with 2.0 app, for example).

test/AspNetCoreModule.TestSites.Standard/TestMiddleWareBeforeIISMiddleWare.cs
Tratcher: Oh, that's what the PathBase and UsePathBase was for.

jhkimnew: @Tratcher Does that mean we don't use the "UsePathBase"? In debugging, I noticed that the Request.PathBase was empty unlike Request.Path and I was wondering that happens because we did not use the flag in aspnetcore.

jhkimnew: @Tratcher Ah, I found UsePathBase on https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/UsePathBaseExtensions.cs.

jkotalik: Yeah that one :) You can specify the pathbase for an app and run that.

jkotalik (approved):

PR #234, opened 2017-11-06
Disable tests for continuous integration. 

jkotalik:
[This change](https://github.com/aspnet/AspNetCoreModule/commit/192a403b9a782b7b692c9c35a48c0262e8a3ad8b) cause tests to be enabled on CI. I'm going to revert some of those change for now. @jhkimnew can you investigate what is wrong with this check?
/cc @ryanbrandenburg this should fix the build.

### DIFF
--- a/korebuild-lock.txt
+++ b/korebuild-lock.txt
@@ -1,0 +1,2 @@
+version:2.1.0-preview1-15552
+commithash:4af2f444ec8cba0faa866dd867370e6aa6df69ea
--- a/test/AspNetCoreModule.Test/FunctionalTestHelper.cs
+++ b/test/AspNetCoreModule.Test/FunctionalTestHelper.cs
@@ -29,7 +29,7 @@ namespace AspNetCoreModule.Test
         public ANCMTestFlags(string attributeValue)
         {
             _attributeValue = attributeValue.ToString();
-        }
+        }   
 
         public bool IsMet
         {
@@ -47,6 +47,7 @@ namespace AspNetCoreModule.Test
                     AdditionalInfo = _attributeValue + " is not belong to the given global test context(" + InitializeTestMachine.GlobalTestFlags + ")";
                     return false;
                 }
+
                 return true;
             }
         }
--- a/test/AspNetCoreModule.Test/FunctionalTestHelper.cs
+++ b/test/AspNetCoreModule.Test/FunctionalTestHelper.cs
@@ -35,8 +35,7 @@ namespace AspNetCoreModule.Test
         {
             get
             {
-                if (_attributeValue == TestFlags.SkipTest
-                    || TestFlags.Enabled(TestFlags.SkipTest))
+                if (_attributeValue == TestFlags.SkipTest)
                 {
                     AdditionalInfo = TestFlags.SkipTest + " is set";
                     return false;
--- a/test/AspNetCoreModule.Test/FunctionalTestHelper.cs
+++ b/test/AspNetCoreModule.Test/FunctionalTestHelper.cs
@@ -29,7 +29,7 @@ namespace AspNetCoreModule.Test
         public ANCMTestFlags(string attributeValue)
         {
             _attributeValue = attributeValue.ToString();
-        }   
+        }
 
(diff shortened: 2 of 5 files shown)


test/AspNetCoreModule.Test/FunctionalTestHelper.cs
```diff
@@ -35,14 +35,14 @@ public bool IsMet
         {
             get
             {
```
jkotalik, line 37: These are used for an attribute to determine whether the tests is enabled or not.

korebuild-lock.txt
```diff
@@ -0,0 +1,2 @@
+version:2.1.0-preview1-15552
```
natemcmaster, line 1: 👍

jkotalik:
@natemcmaster does appveyor run in an administrator console?

natemcmaster:
Not sure. I'd have to google it

jkotalik:
Alright, I updated it to strictly skip all ANCM tests. @jhkimnew this good for now?

test/AspNetCoreModule.Test/FunctionalTestHelper.cs
```diff
@@ -35,7 +35,8 @@ public bool IsMet
...
-                if (TestFlags.Enabled(TestFlags.SkipTest))
+                if (_attributeValue == TestFlags.SkipTest
```
jhkimnew: In this condition, test are always got skipped, which is a breaking change to our test pass. So, please update this in order to make some way not to skip whole tests.

jkotalik:
@natemcmaster it seems that both AppVeyor and our ANCM test machine run in an administrator console. I will disable all tests on CI for now until we figure out why it is failing. @jhkimnew I made it so you can just change one line to enable tests for now.

test/AspNetCoreModule.Test/FunctionalTestHelper.cs
```diff
@@ -35,7 +35,7 @@ public bool IsMet
         {
             get
             {
-                if (TestFlags.Enabled(TestFlags.SkipTest))
+                if (_attributeValue == TestFlags.SkipTest)
```
jhkimnew, line 38: @jkotalik I wanted to allow we can set multiple values for the _attributeValue with ";" delimiter such as "RunAsAdministrator;SkipTest". In that case, we can't use "==" here.

jkotalik: I disagree with allowing that for attributes. What else would we allow to be set? Anyways the check here https://github.com/aspnet/AspNetCoreModule/blob/dev/test/AspNetCoreModule.Test/FunctionalTestHelper.cs#L44 is also incorrect then. I am keeping this consistent with what is already there.

jhkimnew: For example, the websocket functional tests will be only applicable only when websocket iis module installed because the websocket module is not available on downlevel OS such as Win7.
In that case, the test case should be set with "RunAsAdministrator;WebSocketModuleAvailable" supposing the test requires the "RunAsAdministrator" environment.

Tratcher: No, we have OS version rules for our WebSocket tests.

jhkimnew: No we can't use OS version, because, even in the latest OS, the IIS websocket module is not enabled by default.

Tratcher: Do you have negative tests for how it fails when the module is missing? If not then make installing the module a system prerequisite so you don't have to worry about it in the test. Our tests have several system prereqs.

jkotalik: Let's handle this in another PR if we want to support this kind of behavior.

jhkimnew (approved):
@jkotalik 
Okay, let's decide how to handle the attribute value later. 
Approved now.

PR #236, opened 2017-11-07
Added StartupWithIStartupFilter

jhkimnew:
Added StartupWithIStartupFilter class in order to apply it only for the test that requires disabling the graceful shutdown feature, not for the default test case.

### DIFF
--- a/test/AspNetCoreModule.Test/FunctionalTestHelper.cs
+++ b/test/AspNetCoreModule.Test/FunctionalTestHelper.cs
@@ -718,6 +718,7 @@ namespace AspNetCoreModule.Test
                     if (!isGraceFullShutdownEnabled)
                     {
                         iisConfig.SetANCMConfig(testSite.SiteName, testSite.AspNetCoreApp.Name, "environmentVariable", new string[] { "GracefulShutdown", "disabled" });
+                        iisConfig.SetANCMConfig(testSite.SiteName, testSite.AspNetCoreApp.Name, "environmentVariable", new string[] { "ANCMTestStartupClassName", "StartupWithIStartupFilter" });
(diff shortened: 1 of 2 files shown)


test/AspNetCoreModule.Test/FunctionalTestHelper.cs
```diff
@@ -718,6 +718,7 @@ public enum ReturnValueType
                     if (!isGraceFullShutdownEnabled)
                     {
                         iisConfig.SetANCMConfig(testSite.SiteName, testSite.AspNetCoreApp.Name, "environmentVariable", new string[] { "GracefulShutdown", "disabled" });
+                        iisConfig.SetANCMConfig(testSite.SiteName, testSite.AspNetCoreApp.Name, "environmentVariable", new string[] { "ANCMTestStartupClassName", "StartupWithIStartupFilter" });
```
Tratcher: StartupWithShutdownDisabled?

jhkimnew: I agree. I renamed the class name.

test/AspNetCoreModule.TestSites.Standard/Program.cs
```diff
@@ -98,6 +93,18 @@ public static void Main(string[] args)
                     })
                     .UseConfiguration(config)
                     .UseStartup<Startup>();
+                }
+                else
+                {
+                    throw new System.Exception("Invalid startup class name : " + startUpClassString);
```
jkotalik: Nit "throw new Exception();

jkotalik (approved):
Small nit but looks good to me

PR #245, opened 2017-11-13
Copy aspnetcore.dll to contentfiles and re-add old nuget locations.

jkotalik:


shirhatti:
Why the change? 😕

davidfowl:
Just update the package in hosting to look for old and new locations. When the change fully propages then delete the code that uses the old locations.

jkotalik:
Full framework collapses all dll references into the same folder. So if we had x64/aspnetcore.dll, it will just have aspnetcore.dll in the folder. This will add aspnetcore.dll to two locations. Effectively what @davidfowl said.

Tratcher:
All of the tests that use this are currently disabled, there's no need for fancy staging.

davidfowl:
In the future I recommend fancy staging. It’ll reduce the stress in your life and will give you practice making cross cutting changes.

nuget/AspNetCore.nuspec
```diff
@@ -13,10 +13,15 @@
...
     <language>en-US</language>
     <tags>Microsoft.AspNetCore.AspNetCoreModule</tags>
+    <contentFiles>
+      <files include="any/any/*/*.dll" buildAction="None" copyToOutput="true" flatten="false" />
+    </contentFiles>
   </metadata>
   <files>
-    <file src="..\artifacts\build\AspNetCore\bin\Release\Win32\aspnetcore.dll" target="runtimes\win7\native\x86\aspnetcore.dll" />
-    <file src="..\artifacts\build\AspNetCore\bin\Release\x64\aspnetcore.dll" target="runtimes\win7\native\x64\aspnetcore.dll" />
+    <file src="..\artifacts\build\AspNetCore\bin\Release\Win32\aspnetcore.dll" target="runtimes\win7\native\aspnetcore_x86.dll" />
```
analogrelay: If we were staging this change (which as @Tratcher says, may be unnecessary), this would be wrong wouldn't it? The old code is looking for aspnetcore.dll, not **aspnetcore_x86.dll**.

analogrelay: Ah, I think I see. There's a matching change in Hosting (https://github.com/aspnet/Hosting/pull/1268) ? If that's the case, you're not really "staging" this change anyway, so we can just go all the way to the `contentFiles` model.

Tratcher (approved):

PR #248, opened 2017-11-16
fix build issue

pan-wang:
fix the build flags to use /MT instead of /MD


AspNetCoreModule.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
```
jkotalik, line 1: Revert these changes to the sln file.

jkotalik (approved):
LGTM after feedback.

PR #257, opened 2017-11-21
Adds In-Process support for shimmed module.

jkotalik:
Needed a few changes in managed but works when loading from inetsrv.


jkotalik:
Offline approval from @pan-wang

Issue #263, opened 2017-11-30
AspNetCoreModuleX64Location and AspNetCoreModuleX86Location are wrong


pakrym:
During https://github.com/aspnet/AspNetCoreModule/commit/c12c938d4a69b08f7ea7a133f5139882c4f8caca `.props` wasn't updated:

https://github.com/aspnet/AspNetCoreModule/blob/dev/nuget/Microsoft.AspNetCore.AspNetCoreModule.props#L4

/cc @jkotalik

jkotalik:
#264

PR #264, opened 2017-11-30
Update AspNetCoreModuleX64Location and X86. 

jkotalik:
For #263


pakrym (approved):
Don't forget to update universe when build passes.

PR #266, opened 2017-12-04
Use aspnetcore-dev

ryanbrandenburg:


build/dependencies.props
```diff
@@ -3,18 +3,18 @@
...
   </PropertyGroup>
   <PropertyGroup Label="Package Versions">
-    <InternalAspNetCoreSdkPackageVersion>2.1.0-preview1-15549</InternalAspNetCoreSdkPackageVersion>
-    <MicrosoftAspNetCoreAllPackageVersion>2.0.0</MicrosoftAspNetCoreAllPackageVersion>
-    <MicrosoftAspNetCoreTestingPackageVersion>2.1.0-preview1-27424</MicrosoftAspNetCoreTestingPackageVersion>
-    <MicrosoftExtensionsLoggingConsolePackageVersion>2.0.0</MicrosoftExtensionsLoggingConsolePackageVersion>
-    <MicrosoftExtensionsLoggingDebugPackageVersion>2.0.0</MicrosoftExtensionsLoggingDebugPackageVersion>
-    <MicrosoftExtensionsLoggingPackageVersion>2.0.0</MicrosoftExtensionsLoggingPackageVersion>
+    <InternalAspNetCoreSdkPackageVersion>2.1.0-preview1-15620</InternalAspNetCoreSdkPackageVersion>
+    <MicrosoftAspNetCoreAllPackageVersion>2.1.0-preview1-27737</MicrosoftAspNetCoreAllPackageVersion>
```
Tratcher: Revert. These intentionally use old versions for regression tests. This repo is also still outside of universe so we don't want circular dependencies.

NuGet.config
```diff
@@ -2,8 +2,6 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
-    <add key="AspNetCoreTools" value="https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json" />
-    <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <!-- Restore sources should be defined in build/sources.props. -->
```
muratg, line 5: Is this file going to be standardized across repos?

jkotalik: What is the point in keeping this file then?

muratg: Yeah that was going to be my next question. Perhaps making sure that `<clear/>` is in effect?

pranavkm: It doesn't. @natemcmaster had some explanation for this in the PR that we originally did this. @natemcmaster ?

pranavkm (approved):

build/dependencies.props
ryanbrandenburg: Some of these versions aren't available without aspnetcore-ci-dev (`Microsoft.AspNetCore.Testing` version `2.1.0-preview1-27424` for example). It should be fine to update those right?

NuGet.config
natemcmaster: The purpose of keeping it around is mostly as a safety check. We want to make sure we're not leaking a dependency to some other feed. If a dev adds a new MSBuild project but the import to sources.props fails, we want restore to fail.

jkotalik (approved):

PR #269, opened 2017-12-06
ANCM Shim/RequestHandler Refactor

jkotalik:
Will be commenting throughout the PR with updates/description.

Items that still need to be cleaned up (will make issues on these):
- Logging. Transfering to use a new ETW logging style.
- Verification. 
- Updating tests to react to shim/requesthandler split.
- Any TODOs.

Until all bits are stabalized, we will not update the version of ANCM in Universe.


AspNetCoreModule.sln
```diff
@@ -34,18 +39,28 @@ Global
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
 		{439824F9-1455-4CC4-BD79-B44FA0A16552}.Debug|Any CPU.ActiveCfg = Debug|Win32
-		{439824F9-1455-4CC4-BD79-B44FA0A16552}.Debug|Win32.ActiveCfg = Release|x64
-		{439824F9-1455-4CC4-BD79-B44FA0A16552}.Debug|Win32.Build.0 = Release|x64
+		{439824F9-1455-4CC4-BD79-B44FA0A16552}.Debug|Win32.ActiveCfg = Debug|Win32
```
jkotalik: Update these build configs to all be the same.

src/AspNetCore/Inc/applicationinfo.h
```diff
@@ -0,0 +1,218 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
```
jkotalik, line 2: This is effectively a record for what applications are currently created.

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -0,0 +1,330 @@
...
+    APPLICATION* pApplication = NULL;
+    STACK_STRU(struFileName, 300);  // >MAX_PATH
+    STRU        hostFxrDllLocation;
+
+    if (m_pApplication != NULL)
+    {
+        goto Finished;
+    }
+
+    FindRequestHandlerAssembly();
```
jkotalik: The requesthandler.dll will have two functions exposed on it; CreateApplication (on a per application basis) and CreateRequestHandler (per request basis). Calling FindRequestHandlerAssembly will call into hostfxr to find it.

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -0,0 +1,330 @@
...
+    {
+        AcquireSRWLockExclusive(&g_srwLock);
+        fLocked = TRUE;
+        if (g_fAspnetcoreRHAssemblyLoaded)
+        {
+            goto Finished;
+        }
+
+        // load assembly and create the application
+        if (m_pConfiguration->QueryHostingModel() == APP_HOSTING_MODEL::HOSTING_IN_PROCESS)
```
jkotalik, line 223: This logic will change based on the changes to hostfxr. The logic (for both in proc and out of proc) will be:
if hostfxrpath exists and hostfxr has the ability to find the requesthandler
   use hostfxr to find the request handler
otherwise if we are running out of process
   use the global requesthandler

Another issue is bitness for inprocess and out of process. If we are running a 32-bit worker process and find a 64 bit hostfxr, out of process can still work, but in-process cannot. However, if we want out-of-process to be loaded the same way we do in process, we would need a hostfxr.dll to be the same bitness as the worker process. We will need to discuss the finer details here.

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -0,0 +1,330 @@
...
+    {
+        ReleaseSRWLockExclusive(&g_srwLock);
+    }
+    return hr;
+}
+
+HRESULT
+APPLICATION_INFO::FindNativeAssemblyFromInetsrv(STRU* struFilename)
+{
+    HRESULT hr = S_OK;
```
jkotalik: Eventually will only be used for backwards compatibility.

src/AspNetCore/Src/applicationmanager.cxx
```diff
@@ -128,21 +125,27 @@ APPLICATION_MANAGER::GetApplication(
...
+    if (FAILED(hr))
+    {
         if (fDuplicatedInProcessApp)
         {
             if (SUCCEEDED(strEventMsg.SafeSnwprintf(
                 ASPNETCORE_EVENT_DUPLICATED_INPROCESS_APP_MSG,
                 pszApplicationId)))
             {
-                apsz[0] = strEventMsg.QueryStr();
+                /*apsz[0] = strEventMsg.QueryStr();
```
jkotalik, line 156: We will be updating logging in a separate PR.

src/AspNetCore/Src/inprocessstoredcontext.cxx
```diff
@@ -1,113 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the MIT License. See License.txt in the project root for license information.
-
-#include "precomp.hxx"
-
-IN_PROCESS_STORED_CONTEXT::IN_PROCESS_STORED_CONTEXT(
```
jkotalik, line 6: Using IN_PROCESS_HANDLER instead of stored context.

src/AspNetCore/Src/proxymodule.cxx
```diff
@@ -41,87 +41,139 @@ Return value:
...
+        retVal = RQ_NOTIFICATION_FINISH_REQUEST;
+        goto Finished;
+    }
+    else
+    {
+        // make sure assmebly is loaded and application is created
+        APPLICATION* pApplication = NULL;
+        STACK_STRU(struFileName, 256);
+
+        hr = pApplicationInfo->EnsureApplicationCreated();
```
jkotalik: EnsureApplicationCreated will load the request handler if need be.

src/CommonLib/CommonLib.vcxproj
```diff
@@ -0,0 +1,204 @@
+<?xml version="1.0" encoding="utf-8"?>
```
jkotalik, line 1: Common lib is for any shared components between AspNetCore and the RequestHandler projects.

jkotalik (commented):
A bit of introduction for some of the inprocess stuff.

pan-wang:
refactor ANCM to support In-process

PR #274, opened 2017-12-13
Adds logic to call into hostfxr for finding native dll path.

jkotalik:
- We still default to the global request handler if we couldn't find it. We still need to await hostfxr changes.
- Creates a new nuget package which contains the request handler. We still want to keep the old aspnetcore nuget package for testing.


src/CommonLib/hostfxr_utility.cpp
```diff
@@ -281,6 +281,7 @@ HOSTFXR_UTILITY::GetPortableHostfxrParameters(
     if (hFileHandle == INVALID_HANDLE_VALUE)
     {
         hr = ERROR_FILE_INVALID;
+        CloseHandle(hFileHandle);
```
pan-wang: you cannot close a handle which has INVALID_HANDLE_VALUE. The handle was not created successfully yet.

src/CommonLib/hostfxr_utility.h
```diff
@@ -2,6 +2,9 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 #pragma once
+
+#define DOTNET_HOME_ENV_STR "DOTNET_HOME"
```
shirhatti: I'm assuming we're no longer using this?

shirhatti:
src/CommonLib/stdafx.h is a binary file? 😕

jkotalik:
Yeah no idea why it thinks that.

shirhatti:
Did you verify that cmdline parameters passed via the `arguments` attribute still work?

jkotalik:
I just realized that command line args must be passed space delimited to hostfxr. I'll make the fix here soon.

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -330,3 +321,115 @@ APPLICATION_INFO::FindNativeAssemblyFromGlobalLocation(STRU* struFilename)
...
+
+        if (FAILED(struNativeDllLocation.Append(g_pwzAspnetcoreRequestHandlerName)))
+        {
+            goto Finished;
+        }
+
+        hNativeRequestHandler = UTILITY::CheckIfFileExists(&struNativeDllLocation);
+        if (hNativeRequestHandler != INVALID_HANDLE_VALUE)
+        {
+            struFilename->Copy(struNativeDllLocation);
```
pan-wang: check the hresult of copy

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -330,3 +321,115 @@ APPLICATION_INFO::FindNativeAssemblyFromGlobalLocation(STRU* struFilename)
...
+
+        if (!struNativeDllLocation.EndsWith(L"\\"))
+        {
+            if (FAILED(hr = struNativeDllLocation.Append(L"\\")))
+            {
+                goto Finished;
+            }
+        }
+
+        if (FAILED(struNativeDllLocation.Append(g_pwzAspnetcoreRequestHandlerName)))
```
pan-wang: set the hr value from the append

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -330,3 +321,115 @@ APPLICATION_INFO::FindNativeAssemblyFromGlobalLocation(STRU* struFilename)
...
+        GetProcAddress(hmHostFxrDll, "hostfxr_get_native_search_directories");
+
+    if (pFnHostFxrSearchDirectories == NULL)
+    {
+        // Host fxr version is incorrect
+        hr = E_FAIL;
+        goto Finished;
+    }
+
+    intHostFxrExitCode = pFnHostFxrSearchDirectories(
```
pan-wang: what would happen if buffer is too small?

src/CommonLib/aspnetcoreconfig.h
```diff
@@ -283,4 +291,7 @@ class ASPNETCORE_CONFIG : IHttpStoredContext
     BOOL                   m_fIsStandAloneApplication;
     APP_HOSTING_MODEL      m_hostingModel;
     ENVIRONMENT_VAR_HASH*  m_pEnvironmentVariables;
+    STRU                   m_struHostFxrLocation;
+    PCWSTR*                m_pcwArguments;
```
pan-wang: it is better to rename it as m_ppStrArguments as it points to a array of string pointers

src/CommonLib/aspnetcoreconfig.h
```diff
@@ -215,18 +215,28 @@ class ASPNETCORE_CONFIG : IHttpStoredContext
...
+    )
     {
-        return &m_struHostFxrPath;
+        return &m_struHostFxrLocation;
     }
 
-    BOOL
-    QueryIsStandAloneApplication(
+    DWORD*
+    QueryHostFxrArgCount(
```
pan-wang: theoretically this counter can be out of sync with the string array. Maybe you can try dynamically calculate the value sizeof(m_pcwArguments)/sizeof(m_pcwArguments[0]) with null check

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -54,160 +24,160 @@ HOSTFXR_UTILITY::FindHostFxrDll(
...
+    HANDLE              hFileHandle = INVALID_HANDLE_VALUE;
+    STRU                struHostfxrPath;
+    STRU                struExePath;
+    STRU                struDllPath;
+    STRU                struArguments;
+    DWORD               dwPosition;
 
-    // Get the full path to the exe and check if it exists
-    if (FAILED(hr = UTILITY::ConvertPathToFullPath(L"\\hostfxr.dll",
+    pConfig->QueryHostFxrLocation()->Copy(struHostfxrPath);
```
pan-wang: the copy may fail. please do a check

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -54,160 +24,160 @@ HOSTFXR_UTILITY::FindHostFxrDll(
...
 
-    hFileHandle = CreateFile(struHostfxrPath->QueryStr(),
-            GENERIC_READ,
-            FILE_SHARE_READ,
-            &saAttr,
-            OPEN_EXISTING,
-            FILE_ATTRIBUTE_NORMAL,
-            NULL);
+    if (FAILED(hr = struDllPath.SyncWithBuffer())
+        || FAILED(hr = struDllPath.Append(L".dll")))
```
pan-wang: move the || to the end of previous line

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -54,160 +24,160 @@ HOSTFXR_UTILITY::FindHostFxrDll(
...
     {
-        // Treat access isseu as File not found
+        // Treat access issue as File not found
         hr = ERROR_FILE_NOT_FOUND;
         goto Finished;
     }
-    else
+
+    if (FAILED(hr = struArguments.Copy(struDllPath))
+        || FAILED(hr = struArguments.Append(L" "))
```
pan-wang: move "||" to the end of previous line

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -54,160 +24,160 @@ HOSTFXR_UTILITY::FindHostFxrDll(
...
+        || struArguments.Append(pConfig->QueryArguments()))
```
pan-wang: missing  FAILED check and hr assignment

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -54,160 +24,160 @@ HOSTFXR_UTILITY::FindHostFxrDll(
...
-    BOOL                        fFound = FALSE;
     HANDLE                      hFileHandle = INVALID_HANDLE_VALUE;
-    SECURITY_ATTRIBUTES         saAttr;
     std::vector<std::wstring>   vVersionFolders;
+    DWORD                       dwPosition;
+    DWORD                       dwLength;
+    WCHAR                       pszDotnetLocation[MAX_PATH];
 
-    if (FAILED(hr))
+    if ((hFileHandle = UTILITY::CheckIfFileExists(pConfig->QueryProcessPath())) != INVALID_HANDLE_VALUE)
```
pan-wang: it's better to put the assignment in one line and if check in another line

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -54,160 +24,160 @@ HOSTFXR_UTILITY::FindHostFxrDll(
...
+            hr = GetStandaloneHostfxrParameters(pConfig);
             goto Finished;
         }
-
-        hr = strDotnetExeLocation.Append(pszDotnetExeString);
-        if (FAILED(hr))
+        else
         {
-            goto Finished;
+            UTILITY::ConvertPathToFullPath(pConfig->QueryProcessPath()->QueryStr(), pConfig->QueryApplicationPath()->QueryStr(), &strDotnetExeLocation);
```
pan-wang: do return value check

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -54,160 +24,160 @@ HOSTFXR_UTILITY::FindHostFxrDll(
...
         }
+    }
+    else if ((dwLength = SearchPath(NULL, L"dotnet", L".exe", MAX_PATH, pszDotnetLocation, NULL)) == 0)
```
pan-wang: please handle small buffer size issue and do retry. Otherwise you may block long path scenario

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -54,160 +24,160 @@ HOSTFXR_UTILITY::FindHostFxrDll(
...
+    {
+        hr = E_FAIL;
```
pan-wang: use GetLastError() instead of E_FAIL

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -235,25 +205,86 @@ HOSTFXR_UTILITY::GetPortableHostfxrLocation(
...
+        hr = ERROR_FILE_INVALID;
+        goto Finished;
+    }
+    
+    if (FAILED(hr = GetArguments(pConfig->QueryArguments(), &strDotnetExeLocation, pConfig)))
     {
         goto Finished;
+    }
 
+    if (FAILED(pConfig->QueryHostFxrLocation()->Copy(struHostFxrPath)))
```
pan-wang: missing hr assignment

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -235,25 +205,86 @@ HOSTFXR_UTILITY::GetPortableHostfxrLocation(
...
+    STRU* pstruExePath, 
+    ASPNETCORE_CONFIG* pConfig
+)
+{
+    HRESULT     hr = S_OK;
+    INT         argc = 0;
+    PCWSTR*     argv = NULL;
+    LPWSTR*     pwzArgs = NULL;
+
+    pwzArgs = CommandLineToArgvW(struArgumentsFromConfig->QueryStr(), &argc);
```
pan-wang, line 280: please do null check on returned pwzArgs

src/CommonLib/hostfxr_utility.cpp
```diff
@@ -235,25 +205,86 @@ HOSTFXR_UTILITY::GetPortableHostfxrLocation(
...
 
+    argv[0] = SysAllocString(pstruExePath->QueryStr());
+    argv[1] = SysAllocString(L"exec");
+
+    for (INT i = 0; i < argc; i++)
+    {
+        argv[i + 2] = SysAllocString(pwzArgs[i]);
+    }
+
+    *pConfig->QueryHostFxrArgCount() = argc + 2;
```
pan-wang: It's not a good practice to do change the returned value of Query function. You should have a Set function. Query function should return const value

src/AspNetCore/Src/applicationinfo.cpp
muratg: Also, with these different failure modes, do we want to log causes as well? I'd like it if it was super straightforward to diagnose common mistakes and recover from them.

src/CommonLib/utility.cxx
```diff
@@ -606,4 +567,25 @@ UTILITY::FindDotNetFolders(
     } while (FindNextFileW(handle, &data));
 
     FindClose(handle);
-}
\ No newline at end of file
+}
+
+HANDLE
+UTILITY::CheckIfFileExists(STRU* struFile)
+{
```
pan-wang: the name of this function does not match its functionality. If you intent to check file exists, the return value should be bool and you should also handle the scenario of lack access permission

src/CommonLib/utility.cxx
```diff
@@ -606,4 +567,25 @@ UTILITY::FindDotNetFolders(
     } while (FindNextFileW(handle, &data));
 
     FindClose(handle);
-}
\ No newline at end of file
+}
+
+HANDLE
+UTILITY::CheckIfFileExists(STRU* struFile)
```
pan-wang: the input can be "_In_ PCWSTR pszFilePath" instead of STRU* as you do not use STRU functions

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -495,6 +391,11 @@ IN_PROCESS_APPLICATION::LoadManagedApplication
     {
         // Core CLR has already been loaded.
         // Cannot load more than once even there was a failure
+        if (m_fLoadManagedAppError)
+        {
+            hr = E_FAIL;
```
pan-wang: can we give some meaningful hresult value here

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -505,6 +406,11 @@ IN_PROCESS_APPLICATION::LoadManagedApplication
     fLocked = TRUE;
     if (m_fManagedAppLoaded || m_fLoadManagedAppError)
     {
+        if (m_fLoadManagedAppError)
+        {
+            hr = E_FAIL;
```
pan-wang: same suggestion as the above one

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -635,150 +544,13 @@ IN_PROCESS_APPLICATION::ExecuteApplication(
...
-    hr = strDotnetFolderLocation.Append(L"\\hostfxr.dll");
-    if (FAILED(hr))
-    {
-        goto Finished;
-    }
-
-    hModule = LoadLibraryW(strDotnetFolderLocation.QueryStr());
+    // should be a redudant call here, but we will be safe and call it twice.
+    // TODO AV here on m_pHostFxrParameters being null
+    hModule = LoadLibraryW(m_pConfig->QueryHostFxrLocation()->QueryStr());
```
pan-wang: it seems this Query function does not return the location. IT returns the full path of the file. Should we rename it?

src/RequestHandler/inprocess/inprocessapplication.cpp
```diff
@@ -862,14 +627,15 @@ IN_PROCESS_APPLICATION::ExecuteApplication(
 // Calls hostfxr_main with the hostfxr and application as arguments.
 // Method should be called with only 
 // Need to have __try / __except in methods that require unwinding.
+// Note, this will not 
 // 
 HRESULT
-IN_PROCESS_APPLICATION::RunDotnetApplication(PCWSTR* argv, hostfxr_main_fn pProc)
+IN_PROCESS_APPLICATION::RunDotnetApplication(CONST DWORD argc, PCWSTR* argv, hostfxr_main_fn pProc)
```
pan-wang: do you want to trust the passed in argc matching with the counter of argv

src/AspNetCore/Src/applicationinfo.cpp
pan-wang: we can log at one place instead of multiple locations. However, the hr value must be meaningful so that we easily identify the root cause based on the hr value

jkotalik:
I'll address the feedback here then move the PR to IISIntegration.

src/AspNetCore/Src/applicationinfo.cpp
jkotalik: For sure. I think we are batching all logging related things into the same PR.

src/CommonLib/utility.cxx
jkotalik: > you should also handle the scenario of lack access permission

When you call CreateFile with a lack of access permission, wouldn't that still return INVALID_HANDLE_VALUE?

src/RequestHandler/inprocess/inprocessapplication.cpp
jkotalik: I think it's a valid assumption.

src/CommonLib/utility.cxx
```diff
@@ -606,4 +567,34 @@ UTILITY::FindDotNetFolders(
...
+}
+
+BOOL
+UTILITY::CheckIfFileExists(
+    _In_ PCWSTR pszFilePath
+)
+{
+    HANDLE              hFileHandle = INVALID_HANDLE_VALUE;
+    SECURITY_ATTRIBUTES saAttr;
+    BOOL                fFileExists;
```
pan-wang: please initialize the value to FALSE

src/CommonLib/utility.cxx
```diff
@@ -606,4 +567,34 @@ UTILITY::FindDotNetFolders(
...
+
+    saAttr.nLength = sizeof(SECURITY_ATTRIBUTES);
+    saAttr.bInheritHandle = TRUE;
+    saAttr.lpSecurityDescriptor = NULL;
+
+    hFileHandle = CreateFile(pszFilePath,
```
pan-wang, line 585: if the file is opened by another thread for writing, CreateFile will fail. You may need to check the last error code to see whether it is permission issue

src/CommonLib/aspnetcoreconfig.h
```diff
@@ -215,18 +215,38 @@ class ASPNETCORE_CONFIG : IHttpStoredContext
...
+        return &m_struHostFxrLocation;
+    }
+
+    VOID
+    SetHostFxrArguments(
+        DWORD dwArgc,
+        PCWSTR* ppStrArguments
+    )
+    {
+        m_dwArgc = dwArgc;
```
pan-wang: there will be memory leak if this function is called more than once. please free m_ppStrArgments first if it is not NULL and then do the assignment

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -330,3 +329,134 @@ APPLICATION_INFO::FindNativeAssemblyFromGlobalLocation(STRU* struFilename)
...
+    STRU        struNativeDllLocation;
+    HMODULE     hmHostFxrDll = NULL;
+    INT         intHostFxrExitCode = 0;
+    INT         intIndex = -1;
+    INT         intPrevIndex = 0;
+    BOOL        fFound = FALSE;
+    DWORD       dwBufferSize = 1024 * 10;
+
+    DBG_ASSERT(struFileName != NULL);
+    DBG_ASSERT(pHostFxrParameters != NULL);
```
pan-wang: where is pHostFxrParameters defined? it is not a passed in parameter

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -330,3 +329,134 @@ APPLICATION_INFO::FindNativeAssemblyFromGlobalLocation(STRU* struFilename)
...
+        goto Finished;
+    }
+
+    hostfxr_get_native_search_directories_fn pFnHostFxrSearchDirectories = (hostfxr_get_native_search_directories_fn)
+        GetProcAddress(hmHostFxrDll, "hostfxr_get_native_search_directories");
+
+    if (pFnHostFxrSearchDirectories == NULL)
+    {
+        // Host fxr version is incorrect
+        hr = E_FAIL;
```
pan-wang: this is mismatch scenario. We need log the error here. You can add a todo here

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -330,3 +329,134 @@ APPLICATION_INFO::FindNativeAssemblyFromGlobalLocation(STRU* struFilename)
...
+    if (!fFound)
+    {
+        hr = E_FAIL;
+        goto Finished;
+    }
+
+Finished:
+    if (FAILED(hr) && hmHostFxrDll != NULL)
+    {
+        FreeLibrary(hmHostFxrDll);
```
pan-wang, line 462: will we do retry (on next request) in case of load hostfxr failure?

src/CommonLib/aspnetcoreconfig.h
```diff
@@ -215,18 +215,38 @@ class ASPNETCORE_CONFIG : IHttpStoredContext
...
+        return m_dwArgc;
+    }
+
+    STRU*
+    QueryHostFxrFullPath(
         VOID
     )
     {
-        return m_fIsStandAloneApplication;
+        return &m_struHostFxrLocation;
```
pan-wang: the return type can be changed tp const PCWSTR i.e, return m_struHostFxrLocation->QueryStr(). This can prevent others from accidently change the stru.

src/CommonLib/utility.cxx
jkotalik: Any reason why I can't use PathFileExists? https://msdn.microsoft.com/en-us/library/windows/desktop/bb773584(v=vs.85).aspx

pan-wang: you can. the limitation is long file path support (which we claimed on Win10) and dependency on Shlwapi.dll.

jkotalik: We already have that dependency on that dll. Considering we are avoiding file path issues already, I will continue to use CreateFile.

pan-wang: then we can use it.

src/CommonLib/aspnetcoreconfig.h
jkotalik: Currently the STRU is modified outside of config. Other pieces of config use this pattern where query returns a pointer to a STRU: https://github.com/aspnet/AspNetCoreModule/blob/dev/src/CommonLib/aspnetcoreconfig.h#L122-L128. I can make this a set/query pattern if you would like.

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -330,3 +329,134 @@ APPLICATION_INFO::FindNativeAssemblyFromGlobalLocation(STRU* struFilename)
...
+    DWORD       dwBufferSize = 1024 * 10;
+
+    DBG_ASSERT(struFileName != NULL);
+
+    hmHostFxrDll = LoadLibraryW(m_pConfiguration->QueryHostFxrFullPath()->QueryStr());
+    
+    if (hmHostFxrDll == NULL)
+    {
+        // Could not load hostfxr
+        goto Finished;
```
pan-wang: the hresult will be S_OK for this case. Is this you expected?

src/CommonLib/aspnetcoreconfig.h
```diff
@@ -215,18 +215,43 @@ class ASPNETCORE_CONFIG : IHttpStoredContext
...
+    {
+        return &m_struHostFxrLocation;
+    }
+
+    VOID
+    SetHostFxrArguments(
+        DWORD dwArgc,
+        PCWSTR* ppStrArguments
+    )
+    {
```
pan-wang, line 258: it will be nice to add debug assert here indicating the contact between caller and callee

pan-wang (commented):
other changes looks good except the LoadLibarayW failure case

jkotalik:
@pan-wang I changed the functions to be set/query. Is your latest comment approval of this PR?

src/AspNetCore/Src/applicationinfo.cpp
```diff
@@ -330,3 +329,137 @@ APPLICATION_INFO::FindNativeAssemblyFromGlobalLocation(STRU* struFilename)
...
+    DWORD       dwBufferSize = 1024 * 10;
+
+    DBG_ASSERT(struFileName != NULL);
+
+    hmHostFxrDll = LoadLibraryW(m_pConfiguration->QueryHostFxrFullPath());
+    
+    if (hmHostFxrDll == NULL)
+    {
+        // Could not load hostfxr
+        hr = GetLastError();
```
pan-wang: last Error is winerror code not hresult. use HresultFromWin32(GetLastError())

pan-wang (approved):
one minor change

pan-wang (approved):

PR #276, opened 2017-12-13
Fixes back compat versioning issue

jkotalik:
When changing the path to where aspnetcore was, I should have incremented the major version number of the AspNetCoreModule package. Issues arise when using older apps (1.0.x and 2.0.x) that are using floating version numbers. For now, we will have to bring back the other locations in the nupkg.


jkotalik:
Will Reopen in IISIntegration

PR #289, opened 2018-02-15
Jhkim/test inprocess

jhkimnew:
Fixing test issues


jhkimnew:
This is for updating test code fix. The solution file is updated, which is okay because we don't use this solution file to build aspnetcore module.

PR #301, opened 2018-03-08
WebSocketClientEXE is updated to make it more reliable.

jhkimnew:
@bangbingsyb I have started updating websocketclientexe to add more test coverage and make it more reliable.


Tratcher:
Why are you still sending PRs to this repo? I thought everything had moved to IISIntegration.

jhkimnew:
@jkotalik  moved only ANCM to IISIntegration, not ANCM test, and that's why I am still using this repo for updating test code.

Tratcher:
We should also move the tests, there's no reason to keep things separate.

jkotalik:
@jhkimnew I would like to eventually have these tests as part of our main build (in IISIntegration). We should outline the steps necessary to make this happen.

PR #310, opened 2018-04-25
Deprecate this repository

natemcmaster:
HTTP 301
Location: https://github.com/aspnet/IISIntegration

All active development should move to https://github.com/aspnet/IISIntegration. We have stopped building this repository on CI, and changes made to it are not included in releases.


natemcmaster:
@Eilon you can disable AppVeyor on this one now. :)

muratg (commented):
@Eilon isn't there a more direct way to deprecate a repo?

muratg (commented):
@pan-wang @jhkimnew FYI. I don't believe you are using this repo, but if you are, let's move things out to IISIntegration or elsewhere instead.

jhkimnew:
@muratg @natemcmaster Currently we use this repository for the legacy test automation program of ANCM module. 
@jkotalik Can you help moving the test project in this repository to IISIntegration so that I can use IISIntegration and we can disable this repository?

jhkimnew:
@shirhatti  @pan-wang @jkotalik One question. As you know this repository is linked to VSTS DevDiv. Can we use the VSTS directly instead of using GitHub for developing our test automation? If so, we don't need to move the test project in this repository to IISIntegration repository.

pan-wang:
If my memory is correct,  we did not move it to iisintegration due to loop denpendency. Moving to VSTS will make the test project private.  I personally prefer to create another repo calling ANCM end2end test repo if possible.

muratg:
Which test is this exactly on this repo? And what's the cyclic dependency?

natemcmaster:
IIRC the cyclical dependency was that the tests relied on the ASP.NET Core shared framework (Microsoft.AspNetCore.App). In our current system, repos that build as a part of producing that shared framework cannot also rely on that shared framework.

An alternative to deleting the code on the dev branch is to archive it on GitHub. That makes it entirely read-only. It would still be good to update the README before doing that to point users to the right location for ongoing development.

Tratcher:
Microsoft.AspNetCore.App isn't required for testing most scenarios, we could remove that dependency.

jkotalik:
There were a few reasons why I didn't move all of the tests to IISIntegration initially:
- Circular dependencies: More than just M.A.App/All. Any references to things above IISIntegration on the stack needs to be removed.
- Test maintainability: It is hard for me to understand what each test is doing
- Timeouts/delays: A normal test run takes ~30 minutes of all of these tests, which is unacceptable in our build system.

I think we can all agree that more test coverage is a good thing, however if this repository is only used for tests now, we should come up with a strategy for porting all of the tests over to IISIntegration. Here is what I propose:

1. Someone (@jhkimnew, @shirhatti, myself, etc.) forks this repository under there own github profile.
2. Delete this repository.
3. Work on porting a single test to IISIntegration.
4. Port the rest of the tests to IISIntegration.

Thoughts?

Tratcher:
Not much point to creating another fork. We don't have to kill this repo immediately. Start the migration and we'll kill this when you're done.

natemcmaster:
It looks like development has continued on this repository. I think we should really consider not having two public repos with the same code, but I don't have the right powers to make this happen, so closing in favor of letting the managers take care of this.

PR #317, opened 2018-06-05
Cleanup script to be run before CI workflow

shhsu:
Sometimes between CI runs, things are not properly cleaned up


tools/ci_ensure_cleanup.ps1
```diff
@@ -0,0 +1,35 @@
...
+        Stop-Process -Id $proc.ProcessId -Force
+    }
+}
+
+# Some failed test would actually cause appHost config file to become corrupted and w3svc would not be able to start
+function EnsureAppHostConfig() {
+    $appConfigLocation = [System.IO.Path]::Combine($env:WinDir, "System32", "inetsrv", "config", "applicationHost.config")
+    if (!(Test-Path $appConfigLocation) -or ((Get-Item $appConfigLocation).Length -eq 0)) {
+        $appConfigBackupLocation = "${appConfigLocation}.ancmtest.masterbackup"
+        Copy-Item -Path $appConfigBackupLocation -Destination $appConfigLocation -Force
```
jhkimnew, line 29: After copying the file, you should double-check the file size is the same to the backup file. 
In some cases, the applicationhost.config file can be locked by other process and the copy-item will fail with throwing some exception error.
So, I'd like to suggest to add a exception handling for the copy-item and restart IIS services to force to unlock the file handle lock if the file size is not matched and try again.

shhsu: Good point. Maybe I should just fail the entire script at this point and block the CI build.
We can come back in and rerun the test or login to CI machine to manually fix the problem.

jhkimnew (approved):
Looks good!!!

PR #319, opened 2018-08-06
Jkotalik/upgrade deps

jkotalik:


test/AspNetCoreModule.Test/AspNetCoreModule.Test.csproj
```diff
@@ -20,11 +20,11 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Testing" Version="2.1.0-preview1-27644" />
```
shhsu, line 23: Does these package needs to be updated in order to work? Last I checked upgrading these packages broke some tests

shhsu (approved):
Thanks for investigating!

test/AspNetCoreModule.Test/AspNetCoreModule.Test.csproj
jkotalik: If that is the case, the tests should be updated. We can revert this if tests start failing.

